PR #1, opened 2020-06-08
Bootstrap schemas + samples validation

bcaudan:
### Schemas

Introduce a single schema for RUM events that try to match one of the event type schema.
Extract common properties to a common schema.

### Validation

Introduce event samples and a validation command to ensure that schema declaration is compatible with actual events.

### Next step

Check with #opensource to open the repository


schemas/_common-schema.json
```diff
@@ -0,0 +1,90 @@
...
+    "type",
+    "date",
+    "application",
+    "session",
+    "view",
+    "_dd"
+  ],
+  "properties": {
+    "type": {
+      "type": "string",
```
xgouchet, line 16: We could add 
```json
"enum": ["view", "action", "resource", "error", "long_task"]
```

to restrict the accepted type values

schemas/_common-schema.json
```diff
@@ -0,0 +1,90 @@
...
+        "id",
+        "type"
+      ],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "UUID of the session"
+        },
+        "type": {
+          "type": "string",
```
xgouchet, line 53: Likewise we could use an `enum` to define the different session types

schemas/action-schema.json
```diff
@@ -0,0 +1,91 @@
...
+    {
+      "required": [
+        "action"
+      ],
+      "properties": {
+        "action": {
+          "type": "object",
+          "description": "Action properties",
+          "required": [
+            "type"
```
mariusc83, line 19: shouldn't we make also the `id` required ?

schemas/_common-schema.json
```diff
@@ -0,0 +1,90 @@
...
+          "description": "Type of the session"
+        }
+      }
+    },
+    "view": {
```
mariusc83, line 59: should this be part of the `common-schema` ?

schemas/action-schema.json
```diff
@@ -0,0 +1,91 @@
...
+          ],
+          "properties": {
+            "type": {
```
mariusc83, line 22: Shouldn't we add an  "enum" : ["tap", "scroll", "swipe"] here ?

schemas/error-schema.json
```diff
@@ -0,0 +1,63 @@
...
+            "resource": {
+              "type": "object",
+              "description": "Resource properties of the error",
+              "required": [
+                "method",
+                "status_code",
+                "url"
+              ],
+              "properties": {
+                "method": {
```
mariusc83, line 45: Maybe an  "enum" : ["GET","POST","PUT"] here will help

schemas/view-schema.json
```diff
@@ -0,0 +1,136 @@
...
+      ],
+      "properties": {
+        "view": {
+          "type": "object",
+          "description": "View properties",
+          "required": [
+            "id",
+            "url",
+            "loading_time",
+            "loading_type",
```
mariusc83: I am not sure we are sending this from Mobile so maybe this should not be mandatory . @xgouchet ?

schemas/_common-schema.json
bcaudan: sounds great, I'll add that for the different types

schemas/action-schema.json
bcaudan: we don't have id for custom action

schemas/_common-schema.json
bcaudan: it is on the common schema, what do you mean?

schemas/_common-schema.json
bcaudan: https://github.com/DataDog/rum-events-format/pull/1#discussion_r436548590

schemas/action-schema.json
bcaudan: https://github.com/DataDog/rum-events-format/pull/1#discussion_r436548590

schemas/error-schema.json
bcaudan: https://github.com/DataDog/rum-events-format/pull/1#discussion_r436548590

schemas/view-schema.json
bcaudan: IIRC, it should be added on the mobile, not sure if it is a prerequisite for the v2 format on the mobile side though

schemas/_common-schema.json
mariusc83: What I wanted to say is that our mobile events don't have this common part (e.g. an Action event does not contain this `view` block) so I am wondering if this should really be part of the `common_schema`

schemas/view-schema.json
xgouchet: The thing is we're working on providing some performance metrics, but not sure what those metrics will be. Chances are they might get their own name as it might not be the "loading time" or loading type

schemas/_common-schema.json
xgouchet: All event do have some info about the current view (namely `view.id` and `view.url`)

mariusc83: Oh...nevermind ...forget what I said :)...we have it also here

schemas/view-schema.json
bcaudan: For consistency between mobile and web, it could be nice to have a similar metric.
I can remove it from required if there are more thoughs to put on your side

xgouchet: In general I too prefer using similar metrics, but if the meaning are too different, it would cause confusion. For now let's make those two non required, until we do know what metrics we'll send and how.

mariusc83 (dismissed):

xgouchet (dismissed):

schemas/resource-schema.json
```diff
@@ -0,0 +1,163 @@
...
+          "properties": {
+            "type": {
+              "type": "string",
+              "description": "Resource type",
+              "enum": ["document", "xhr", "beacon", "fetch", "css", "js", "image", "font", "media", "other"]
+            },
+            "method": {
+              "type": "string",
+              "description": "HTTP method of the resource",
+              "enum": ["POST", "GET", "HEAD", "PUT", "DELETE"]
```
glorieux: No `PATCH`?

schemas/error-schema.json
```diff
@@ -0,0 +1,65 @@
...
+                  "type": "string",
+                  "description": "HTTP method of the resource",
+                  "enum": ["POST", "GET", "HEAD", "PUT", "DELETE"]
```
glorieux: No `PATCH`?

schemas/_common-schema.json
```diff
@@ -0,0 +1,92 @@
...
+  "properties": {
+    "type": {
+      "type": "string",
+      "description": "RUM event type",
+      "enum": ["action", "error", "long_task", "resource", "view"]
+    },
+    "date": {
+      "type": "integer",
+      "description": "Start of the event in ms from epoch"
+    },
```
glorieux: Could we set a minimum?

schemas/_common-schema.json
```diff
@@ -0,0 +1,92 @@
...
+    "application": {
+      "type": "object",
+      "description": "Application properties",
+      "required": [
+        "id"
+      ],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "UUID of the application"
```
glorieux: Could we use a `pattern` to validate the UUID format?

schemas/action-schema.json
```diff
@@ -0,0 +1,93 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://github.com/DataDog/rum-events-format/schemas/action-schema.json",
+  "type": "object",
+  "description": "Schema of all properties of an Action event",
+  "allOf": [
+    {
+      "$ref": "file:schemas/_common-schema.json"
```
xgouchet: Here's a random musing, I can't find in the JsonSchema specification any definition for the `file:` prefix. In all samples I found, referencing another file was done using `"$ref": "folder/another_file.json"` (unless you use an absolute path, eg: `"$ref": "file:///path/to/folder/another_file.json"` ). Where  did you get this from ?

glorieux (dismissed):
LGTM.
We might be able to add a couple of validation rules on URLs.
But it can be added in another iteration.

schemas/action-schema.json
bcaudan: Hum indeed, I probably get that from SO when it did not work OOB with the spec version.
It seems that the correct way to do that with python is to defined a `RefResolver`, cf https://github.com/Julian/jsonschema/issues/98.
I would be in favor of checking if the `file:` approach is working with validator in SDKs repo.
If it does, stick with it.
Otherwise, update it to something like `"$ref": "folder/another_file.json"`, IMO absolute path does not sound great here.
What do you think?

validate.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex
+
+jsonschema -i samples/action.json rum-events-format.json
+jsonschema -i samples/error.json rum-events-format.json
+jsonschema -i samples/long_task.json rum-events-format.json
+jsonschema -i samples/resource.json rum-events-format.json
+jsonschema -i samples/view.json rum-events-format.json
```
BenoitZugmeyer: What about something like:

```bash
#!/usr/bin/env bash

set -e

for sample in samples/*.json; do
  printf "Validating $sample: "
  jsonschema -i $sample rum-events-format.json
  printf "OK\n"
done
```

1. Better feedback
2. Easier to add samples

BenoitZugmeyer (dismissed):
Awesome work. Just a small suggestion for the tooling.

validate.sh
bcaudan: I thought about the loop, I did not want to put too much thoughts on this before getting feedback, I can definitely do that.

bcaudan: with https://github.com/DataDog/rum-events-format/pull/1/commits/7740772ee158f949caaed23b576b6a20ceb3a290:
```
Validating samples/action.json --------------------- ✅
Validating samples/error.json ---------------------- ❌
{u'_dd': {u'format_version': 2}, u'application': {u'id': u'ac8218cf-4...
```

BenoitZugmeyer: Waw :)

BenoitZugmeyer (approved):

schemas/action-schema.json
xgouchet: I think we should follow the specs and use relative paths. The framework we use support them, but doesn't support he `file:` prefix (unless it's an absolute path)

bcaudan: ok, it makes sense, I'll look into it

louisj-dd (approved):

mquentin (approved):

schemas/action-schema.json
bcaudan: fixed in https://github.com/DataDog/rum-events-format/pull/1/commits/f18cf1b091e1d72dc5c900f3ac055dbcaef31c6f

rum-events-format.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum-events-format.json",
+  "type": "object",
+  "description": "Schema of all properties of a RUM event",
+  "anyOf": [
```
xgouchet: Shouldn't we use `oneOf` instead ? 
> To validate against anyOf, the given data must be valid against **any (one or more)** of the given subschemas.
> To validate against oneOf, the given data must be valid against **exactly one** of the given subschemas.

xgouchet (approved):

PR #5, opened 2020-06-19
Add 'BACK' action for mobile apps

xgouchet:


schemas/action-schema.json
```diff
@@ -28,7 +28,7 @@
             "type": {
               "type": "string",
               "description": "Type of the action",
-              "enum": ["custom", "click", "tap", "scroll", "swipe"]
+              "enum": ["custom", "click", "tap", "scroll", "swipe", "back"]
```
bcaudan: on browser side, we don't consider back as an action, it just triggers a new view.
Any reason to promote it as an action?

xgouchet: In a native app, the back button is a button that can be intercepted by app authors to be used as undo, and will not necessarily means "navigating back to the previous view". Most of the time, this action will actually be a navigation and the event won't be sent, but for the few cases out there, it will be detected as a back action.

bcaudan: I see, neat!

bcaudan (approved):

louisj-dd (approved):

PR #7, opened 2020-06-23
Add mobile crash related fields

xgouchet:


mariusc83 (approved):

schemas/action-schema.json
```diff
@@ -67,6 +67,20 @@
...
+            "crash": {
+              "type": "object",
+              "description": "Properties of the crashes of the action",
+              "required": [
+                "count"
+              ],
+              "properties": {
+                "count": {
+                  "type": "integer",
+                  "description": "Number of crashes that occurred on the action",
```
bcaudan, line 79: Can we have multiple crashes on a single view?
I would assume that a single crash kills the application so reopen the application would start a new view.
What is the real behavior?

xgouchet: Yes you're correct, once a view/action has a crash, it can't have more. 
IIRC, the discussion with Nicolas was that we want to mimic the behavior of error_count, and have a numeric instead of boolean value to ease integrating with dashboards.

bcaudan: ok, I get it, thx!

bcaudan (approved):

louisj-dd:
we do not enforce schema validation on backend side, so it's all good to add such field.
Though, at least for the time of the migration, could you add them to the spreadsheet https://docs.google.com/spreadsheets/d/14adhT0u9-k0qolQh3bMLEQanx3Gg6xa5iJkElMzB9xo/edit#gid=0 ?
Also, please tell us if you need a facet on some of these.

louisj-dd (approved):

PR #10, opened 2020-07-03
Add 'usr' and 'connectivity' common attributes

xgouchet:


schemas/_common-schema.json
```diff
@@ -73,6 +73,55 @@
...
+          "type": "string",
+          "description": "Status of the device connectivity",
+          "enum": ["connected", "not_connected", "maybe"]
+        },
+        "interfaces": {
+          "type": "array",
+          "description": "The list of available network interfaces",
+          "items": {
+            "type": "string",
+            "enum": ["wifi", "cellular", "internet", "loopback", "other"]
```
ncreated: 😉 
```suggestion
            "enum": ["wifi", "cellular", "ethernet", "loopback", "other"]
```

schemas/_common-schema.json
```diff
@@ -73,6 +73,55 @@
...
+          }
+        },
+        "cellular_technology": {
+          "type": "string",
+          "description": "The type of a radio technology used for cellular connection"
+        },
+        "cellular_carrier": {
```
ncreated: How about `cellular_carrier_name`?
```suggestion
        "cellular_carrier_name": {
```
This will enable us scoping more carrier information in the future.

ncreated (changes requested):

samples/extended_error.json
```diff
@@ -0,0 +1,43 @@
...
+  "usr": {
+    "id": "4815162342",
+    "name": "John Doe",
+    "email": "john@doe.com"
+  },
+  "connectivity": {
+    "status": "connected",
+    "interfaces": [ "wifi", "cellular", "other"],
+    "cellular_technology": "HSDPA",
+    "cellular_carrier": "AT&T"
```
ncreated: Shouldn't this become:
```suggestion
    "cellular": {
       "technology": "HSDPA",
       "cellular_name": "AT&T"
    },
```
now?

ncreated (approved):
🚀

schemas/_common-schema.json
```diff
@@ -73,6 +73,61 @@
...
+          "type": "string",
+          "description": "Status of the device connectivity",
+          "enum": ["connected", "not_connected", "maybe"]
+        },
+        "interfaces": {
+          "type": "array",
+          "description": "The list of available network interfaces",
+          "items": {
+            "type": "string",
+            "enum": ["wifi", "cellular", "ethernet", "loopback", "other"]
```
bcaudan: fyi browser spec https://wicg.github.io/netinfo

schemas/_common-schema.json
```diff
@@ -73,6 +73,61 @@
...
+      "description": "Device connectivity properties",
+      "required": [
+        "status",
+        "interfaces"
+      ],
+      "properties": {
+        "status": {
+          "type": "string",
+          "description": "Status of the device connectivity",
+          "enum": ["connected", "not_connected", "maybe"]
```
bcaudan, line 105: maybe sounds a bit strange, what cases does it cover?

schemas/_common-schema.json
```diff
@@ -73,6 +73,61 @@
...
+          "description": "Cellular connectivity properties",
+          "properties": {
+            "technology": {
+              "type": "string",
+              "description": "The type of a radio technology used for cellular connection"
+            },
+            "carrier_name": {
+              "type": "string",
+              "description": "The name of the SIM carrier"
+            }
```
bcaudan, line 126: do we really want to provide this level of detail?

schemas/_common-schema.json
ncreated: It comes from [iOS spec](https://developer.apple.com/documentation/network/nwpath/status/requiresconnection) where `requiresConnection` status is available:
> The path is not currently available, but establishing a new connection may activate the path.

schemas/_common-schema.json
xgouchet: From experience, this can prove very usefull. I remember a time where we had a bug on all request only for a specific Mobile Network Operator which modifier the headers of all requests they handled. It took us some time to discriminate users who had this error until we found they were all users using the same carrier.

schemas/_common-schema.json
xgouchet: Updated

schemas/_common-schema.json
bcaudan: ok 🤷‍♂️

schemas/_common-schema.json
bcaudan: where were your interfaces come from? (loopback is gone)

bcaudan (approved):

louisj-dd (approved):

PR #13, opened 2020-09-01
Add resource extra properties for tracing

bcaudan:
Add extra properties to resource to allow: 

- sdk span generation on the backend side
- link to rum resource from trace UI

### DIFF
--- a/samples/resource.json
+++ b/samples/resource.json
@@ -14,6 +14,7 @@
     "url": "https://app.datadoghq.com/rum/explorer?live=1h&query=&tab=view"
   },
   "resource": {
+    "id": "b83a5d82-cdd1-468d-9bc9-3aa9e54d957a",
     "type": "xhr",
     "method": "POST",
     "url": "https://app.datadoghq.com/api/v1/logs-analytics/aggregate?type=rum",
@@ -34,7 +35,6 @@
   },
   "_dd": {
     "format_version": 2,
-    "resource_id": "b83a5d82-cdd1-468d-9bc9-3aa9e54d957a",
     "span_id": "12345",
     "trace_id": "7890"
   }
--- a/samples/resource.json
+++ b/samples/resource.json
@@ -33,7 +33,10 @@
     "id": "ae3a5d82-cdd1-468d-9bc9-3aa9e54d953c"
   },
   "_dd": {
-    "format_version": 2
+    "format_version": 2,
+    "resource_id": "b83a5d82-cdd1-468d-9bc9-3aa9e54d957a",
+    "span_id": "12345",
+    "trace_id": "7890"
   }
 }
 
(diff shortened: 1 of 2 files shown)


xgouchet (approved):

samples/resource.json
```diff
@@ -33,7 +33,10 @@
     "id": "ae3a5d82-cdd1-468d-9bc9-3aa9e54d953c"
   },
   "_dd": {
-    "format_version": 2
+    "format_version": 2,
+    "resource_id": "b83a5d82-cdd1-468d-9bc9-3aa9e54d957a",
```
vlad-mh: specifically for resource ID, shouldn't we be consistent with view, action, etc and have it be `@resource.id`?

bcaudan: It should be indeed more consistent, fine by me.
@xgouchet?

schemas/resource-schema.json
```diff
@@ -191,6 +196,22 @@
               "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
             }
           }
+        },
+        "_dd": {
+          "type": "object",
+          "description": "Internal properties",
+          "properties": {
+            "span_id": {
+              "type": "string",
```
vlad-mh, line 205: just checked, APM defines these as uint64 in their protobuf files: https://github.com/DataDog/dd-go/blob/prod/pb/proto/trace/agent/model/span.proto#L13-L14

does the "integer" type handle 64bit? if so, we can use it instead of string

samples/resource.json
xgouchet: Yep 👍

schemas/resource-schema.json
ncreated: This will work on iOS 👌 , as we use `Int64` for all numeric / integer values in RUM models.

ncreated (approved):

schemas/resource-schema.json
xgouchet: This'll work on our side too

vlad-mh (approved):

schemas/resource-schema.json
```diff
@@ -191,6 +196,20 @@
...
+        "_dd": {
+          "type": "object",
+          "description": "Internal properties",
+          "properties": {
+            "span_id": {
+              "type": "integer",
+              "description": "span identifier in decimal format"
+            },
+            "trace_id": {
+              "type": "integer",
```
BenoitZugmeyer: Is the plan to keep [63 bits integers as trace id](https://github.com/DataDog/browser-sdk/pull/461/files#diff-ad28acad0be8055df13870436c79de04R82-R105) ? If it is, I think the type should be a string, as it may be difficult to represent 63 bits integers values in JS/json. (same for span_id)

schemas/resource-schema.json
bcaudan: Actually, 64 bit integer is not supported on the JS side 😞

BenoitZugmeyer (approved):

PR #14, opened 2020-09-15
Add the first_party attribute for resources

xgouchet:


mariusc83 (approved):

schemas/resource-schema.json
```diff
@@ -180,6 +180,10 @@
                   "minimum": 0
                 }
               }
+            },
+            "first_party": {
```
bcaudan, line 184: why do we need this attribute? how will it be used?

vlad-mh: also, what is a "first party host"? 
the backend already injects `resource.provider.name`, `resource.provider.domain` and `resource.provider.type`. Is it related?

xgouchet: Right now, the backend can't set the category correctly for resources coming from the Mobile SDKs. The categories help filter resources as `first party`, `analytics`,  `social` and so on. The way to determine if a resource is `first party`(that is owned by the org) is by comparing the resource's host with the View Url's. 
This doesn't work for Mobile, because the view "url" is not a url per se but an identifier of the view depending on the framework used. 
The goal with this is that the developers will initialise  the mobile SDKs with a list of "first party" hosts, and we'll set this boolean to true, allowing the intake to categorise those resources as `first party`.

bcaudan: is there a brief or RFC on this topic? I'm wondering why the first party list is configured at SDK level rather than having an app configuration retrieved by the backend.

vlad-mh: same thought, happy to chat

xgouchet: No, there's no RFC, but happy to get on Zoom to discuss this. 
The gist is that it's easier to define a list of first party hosts at the SDK level (and also making it change for different versions of the mobile app), compared to having to create a brand new UI to let the customer define what counts as a first party host for their RUM application

vlad-mh: couldn't we have the SDK set `resource.provider` then if you have more info than the backend?
in the backend, we could then set `resource.provider` only if it isn't already set. wdyt?

bcaudan: >  it's easier to define a list of first party hosts at the SDK level 

Would the customer need to deploy a new version of their app to update this list?


since this new attribute will need to be handled by the intake, a small RFC to initiate the discussion could have been interesting

xgouchet: We talked with Louis earlier about that, sorry about the lack of RFC. I'll create one right now and close this PR until we reach an agreement.

PR #15, opened 2020-09-23
[RUMF-620]: Adds service as optional attribute

glorieux:
Adds service as optional attribute. Will have to be transitioned away from the tags. 

The description was pulled from [the documentation](https://docs.datadoghq.com/logs/log_collection/javascript/?tab=npm).

### DIFF
--- a/schemas/_common-schema.json
+++ b/schemas/_common-schema.json
@@ -30,6 +30,10 @@
         }
       }
     },
+		"service": {
+			"type": "string",
+			"description": "The service name for this application"
+		},
     "session": {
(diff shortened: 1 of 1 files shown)


schemas/_common-schema.json
```diff
@@ -30,6 +30,10 @@
         }
       }
     },
+		"service": {
+			"type": "string",
+			"description": "The service name for this application"
+		},
```
bcaudan: what about using space instead of tab in order to have a consistent formatting?

bcaudan (approved):

PR #17, opened 2020-11-10
[RUMF-774] add Web Vitals related properties

BenoitZugmeyer:
See https://docs.google.com/document/d/1R7xyRWrv9AEsn4xFuUizEBip8C8eZUR5Ajki_RJQXAY/edit for context and implementation details

### DIFF
--- a/schemas/action-schema.json
+++ b/schemas/action-schema.json
@@ -40,6 +40,11 @@
               "description": "Duration in ns to the action is considered loaded",
               "minimum": 0
             },
+            "cumulative_layout_shift": {
+              "type": "number",
+              "description": "Total layout shift score that occured on the action",
+              "minimum": 0
+            },
             "target": {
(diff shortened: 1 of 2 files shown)


schemas/action-schema.json
```diff
@@ -40,6 +40,11 @@
               "description": "Duration in ns to the action is considered loaded",
               "minimum": 0
             },
+            "cumulative_layout_shift": {
```
bcaudan: Since it was not described in product brief, it could be interesting to have some product feedback on that before starting to implement it.

xgouchet (approved):

schemas/action-schema.json
BenoitZugmeyer: I removed it for now

bcaudan (approved):
LGTM

PR #18, opened 2020-11-16
Add the provider attribute for resources

xgouchet:
This PR will allow mobile SDKs to override the Resource's provider attributes beforehand to categorize when possible;

### DIFF
--- a/schemas/resource-schema.json
+++ b/schemas/resource-schema.json
@@ -180,6 +180,25 @@
                   "minimum": 0
                 }
               }
+            },
+            "provider": {
+              "type": "object",
+              "description": "The provider for this resource",
+              "properties": {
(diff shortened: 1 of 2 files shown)


mariusc83 (approved):

schemas/resource-schema.json
```diff
@@ -180,6 +180,25 @@
                   "minimum": 0
                 }
               }
+            },
+            "provider": {
```
bcaudan, line 184: could it be needed for `error.resource` as well?

xgouchet: Valid point indeed.

bcaudan (approved):
LGTM

louisj-dd (approved):

PR #19, opened 2020-12-02
Add read only attribute on fields

bcaudan:
Document which fields we allow customer to modify before events are sent

### DIFF
--- a/schemas/_common-schema.json
+++ b/schemas/_common-schema.json
@@ -78,12 +78,12 @@
         "referrer": {
           "type": "string",
           "description": "URL that linked to the initial view of the page",
-          "readOnly": true
+          "readOnly": false
         },
         "url": {
           "type": "string",
           "description": "URL of the view",
-          "readOnly": true
+          "readOnly": false
         }
       },
       "readOnly": true
--- a/schemas/_common-schema.json
+++ b/schemas/_common-schema.json
@@ -14,7 +14,8 @@
     "date": {
       "type": "integer",
       "description": "Start of the event in ms from epoch",
-      "minimum": 0
+      "minimum": 0,
+      "readOnly": true
     },
     "application": {
       "type": "object",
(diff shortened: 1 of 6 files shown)


schemas/_common-schema.json
```diff
@@ -26,13 +27,16 @@
         "id": {
           "type": "string",
           "description": "UUID of the application",
-          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
+          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+          "readOnly": true
         }
-      }
+      },
+      "readOnly": true
```
xgouchet, line 34: While we're at it, maybe we should add the `additionalProperties` to `false` on the objects, wdyt ?

xgouchet (approved):

schemas/_common-schema.json
bcaudan: hum, the current change don't impact the format validation, so I'd be in favor of not mixing the concerns.

FMU, If we introduce `additionalProperties: false`, we would need to bump format version each time we add a new field, so it may worth to discuss it a bit first.

xgouchet: I didn't think about that indeed, so yeah you're right we shouldn't add the `additionalProperties`

BenoitZugmeyer (approved):
Looks fine. My 2c:
* I have a hard time reading those JSON schema, and I think we could look to generate a more human-readable table for those
* I believe `readOnly` defaults to `false`? we may want to add a linting script to check that all properties have an explicit `readOnly` attribute, so we don't create new writable properties by mistake.

bcaudan:
@BenoitZugmeyer 

> * I have a hard time reading those JSON schema, and I think we could look to generate a more human-readable table for those

What do you have in mind?

> * I believe `readOnly` defaults to `false`? we may want to add a linting script to check that all properties have an explicit `readOnly` attribute, so we don't create new writable properties by mistake.

I think we should setup a ci before adding extra checks, I'll create a task for it

BenoitZugmeyer:
> @BenoitZugmeyer
> 
> > * I have a hard time reading those JSON schema, and I think we could look to generate a more human-readable table for those
> 
> What do you have in mind?
> 

A script generating a markdown array that could looks like https://docs.google.com/spreadsheets/d/14adhT0u9-k0qolQh3bMLEQanx3Gg6xa5iJkElMzB9xo/edit#gid=0 with different columns, like name / type / readonly / description would be nice

bcaudan:
> > @BenoitZugmeyer
> > > * I have a hard time reading those JSON schema, and I think we could look to generate a more human-readable table for those
> > 
> > 
> > What do you have in mind?
> 
> A script generating a markdown array that could looks like https://docs.google.com/spreadsheets/d/14adhT0u9-k0qolQh3bMLEQanx3Gg6xa5iJkElMzB9xo/edit#gid=0 with different columns, like name / type / readonly / description would be nice

Yes, that could be helpful.

louisj-dd (approved):

PR #22, opened 2020-12-10
Add "unknown" to the resource.provider.type enum

BenoitZugmeyer:
This value is set by our backend `ThirdPartyProcessor` if it fails to detect the provider.

### DIFF
--- a/schemas/resource-schema.json
+++ b/schemas/resource-schema.json
@@ -224,7 +224,7 @@
                 "type": {
                   "type": "string",
                   "description": "The type of provider",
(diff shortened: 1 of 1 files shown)


louisj-dd (approved):

BenoitZugmeyer:
On hold until we know if this `unknown` value is expected. The validation step may be happening too late, since the `unknown` value is provided by the log pipeline.

bcaudan:
@BenoitZugmeyer should we close this one?

BenoitZugmeyer:
Yes, let's close this for now.

PR #25, opened 2021-01-08
[RUMF-783] Add view.custom_timings

webNeat:


schemas/view-schema.json
```diff
@@ -99,6 +99,16 @@
...
               "readOnly": true
             },
+            "custom_timings": {
+              "type": "object",
+              "description": "User custom timings of the view",
+              "additionalProperties": {
+                "type": "integer",
+                "minimum": 0,
+                "readOnly": true
+              },
```
ncreated, line 109: I think we've been already discussing this and we decided to not go with `additionalProperties` the same way as we don't use them to model custom context attributes, no? // cc @xgouchet @bcaudan 

Anyway, I think we can deal with this both in Kotlin and Swift by expressing it as `[String: UInt64]` dictionaries.

xgouchet: Well we didn't add this because I didn't know we could constrain the additional properties (i only saw example of `"additionalProperties": true`). 
My guess is that this should actually fix what we wanted and we can make the changes necessary on our side to handle it.

xgouchet (approved):

schemas/view-schema.json
ncreated: Yes, agree - we can incorporate it into our code generators 👍.

ncreated (approved):
🚀

schemas/view-schema.json
bcaudan: IIRC previous discussions was more about having `additionalProperties: false` at every level to constrain our model rather than having `additionalProperties: true` for some specific parent (like context).

bcaudan (approved):

PR #28, opened 2021-02-03
[RUMF-836] add note about timing name restrictions

bcaudan:
cf https://github.com/DataDog/browser-sdk/pull/706

### DIFF
--- a/schemas/view-schema.json
+++ b/schemas/view-schema.json
@@ -107,7 +107,7 @@
             },
             "custom_timings": {
               "type": "object",
-              "description": "User custom timings of the view",
+              "description": "User custom timings of the view, as timing name is used as facet path, it must contain only contains letters, digits, or the characters - _ . @ $",
               "additionalProperties": {
                 "type": "integer",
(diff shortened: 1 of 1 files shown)


xgouchet (approved):

schemas/view-schema.json
```diff
@@ -107,7 +107,7 @@
             },
             "custom_timings": {
               "type": "object",
-              "description": "User custom timings of the view",
+              "description": "User custom timings of the view, as timing name is used as facet path, it must contain only contains letters, digits, or the characters - _ . @ $",
```
BenoitZugmeyer:
```suggestion
              "description": "User custom timings of the view. As timing name is used as facet path, it must only contains letters, digits, or the characters - _ . @ $",
```

bcaudan: thx, I've kept the same phrasing as in the error message in the UI

BenoitZugmeyer (approved):

ncreated (approved):

PR #30, opened 2021-03-03
Update action.target schema with Android fields

rpelliard:
Adds the following fields to the format, that are set for Android actions:
- `action.target.classname`
- `action.target.parent.classname`
- `action.target.resource_id`

### DIFF
--- a/schemas/action-schema.json
+++ b/schemas/action-schema.json
@@ -56,6 +56,27 @@
                   "type": "string",
                   "description": "Target name",
                   "readOnly": false
+                },
+                "classname": {
+                  "type": "string",
+                  "description": "Target class name",
+                  "readOnly": false
+                },
+                "parent": {
+                  "type": "object",
+                  "description": "Parent element of the target",
+                  "properties": {
+                    "classname": {
+                      "type": "string",
+                      "description": "Class name of the parent element",
+                      "readOnly": false
+                    }
+                  }
+                },
+                "resource_id": {
+                  "type": "string",
+                  "description": "identifier for the target resource",
+                  "readOnly": true
                 }
               },
               "readOnly": true


xgouchet (approved):

schemas/action-schema.json
```diff
@@ -56,6 +56,27 @@
...
+                      "description": "Class name of the parent element",
+                      "readOnly": false
+                    }
+                  }
+                },
+                "resource_id": {
+                  "type": "string",
+                  "description": "identifier for the target resource",
+                  "readOnly": true
                 }
```
BenoitZugmeyer, line 80: Is it a RUM resource event id? Maybe it should be `resource.id` like other id fields.

schemas/action-schema.json
```diff
@@ -56,6 +56,27 @@
                   "type": "string",
                   "description": "Target name",
                   "readOnly": false
+                },
+                "classname": {
```
BenoitZugmeyer, line 60: This is a POO "class" name, right? What about using a more abstract term like "component_name", so it could be used for other use-cases?

schemas/action-schema.json
rpelliard: I think the fields are already sent by the SDK, @xgouchet might have more insight on this ?

xgouchet: No those are Android resource ids, nothing to do with RUM Resources. cf https://developer.android.com/guide/topics/resources/providing-resources

BenoitZugmeyer: Then IMO it should be a bit more explicit, like `android_resource_id`

schemas/action-schema.json
xgouchet: That does make sense, we'll need to make some changes but for the best i guess

bcaudan:
are there some details about the need for those properties?
cf https://github.com/DataDog/rum-events-format/pull/30#discussion_r589267210, it could be nice to have a bit of context on what are these properties, why we need them and is there an opportunity to reuse them for other SDKs

schemas/action-schema.json
mariusc83: On the other hand if other teams are not going to use them (e.g. `android_resource_id` will only be used by Android) maybe we should not make them part of the JSON schema.

PR #31, opened 2021-04-21
Add focus types

ThibautGeriz:
Add focus information for browser events:
- in a array for views
- `focus.start_focused` in all other event


schemas/_common-schema.json
```diff
@@ -99,6 +99,21 @@
       },
       "readOnly": true
     },
+    "focus": {
+      "type": "object",
+      "description": "Focius properties",
```
bcaudan:
```suggestion
      "description": "Focus properties",
```

schemas/view-schema.json
```diff
@@ -199,16 +194,44 @@
...
+                    "minimum": 0,
+                    "readOnly": true
+                  },
+                  "duration": {
+                    "type": "integer",
+                    "description": "duration in ns when the user gain focus of the document",
+                    "minimum": 0,
+                    "readOnly": true
+                  },
+                  "currently_focused": {
```
bcaudan, line 225: is it different than `view.focus.has_focused`?

schemas/view-schema.json
```diff
@@ -199,16 +194,44 @@
                 }
               },
               "readOnly": true
+            },
+            "focused_times": {
```
bcaudan, line 205: should it be in the `focus` root?

schemas/_common-schema.json
```diff
@@ -99,6 +99,21 @@
       },
       "readOnly": true
     },
+    "focus": {
```
bcaudan, line 88: since it is not user facing, should it be in `_dd` instead?

schemas/view-schema.json
ThibautGeriz: `view.focus.has_focus` would tell if the view has focus when it opened.
`currently_focused` will give you the change in focus during the life of the view.

schemas/view-schema.json
ThibautGeriz: Maybe, I do not have a string opinion.

schemas/_common-schema.json
ThibautGeriz: I revolved the discussion but in the RFC, the logic was: if it can be useful to us, it can be useful to our client. Hence it should not be private.

schemas/view-schema.json
bcaudan: got it, could it be more accurate to rename `has_focus` in `start_focused` then?

schemas/_common-schema.json
bcaudan: Have not seen this one, ok then

schemas/view-schema.json
bcaudan: I find it a bit weird to have a `focus` group for `has_focus` and not put other focus info in it.

bcaudan (approved):
LGTM

xgouchet (approved):

benoitmeriaux (approved):

PR #32, opened 2021-05-04
Update focus to in_foreground name

ThibautGeriz:


schemas/view-schema.json
```diff
@@ -194,38 +194,26 @@
...
-        "focus": {
-          "type": "object",
-          "description": "Focus properties",
-          "properties": {
-            "focused_times": {
+            },
+            "in_foreground_times": {
               "type": "array",
-              "description": "List of the time the user had the document focused for browser",
+              "description": "List of the period the user had the view in foreground (focused in the browser)",
```
buranmert: is this a browser only property? this can be translated into mobile apps too as they can go background<->foreground

vlad-mh:
```suggestion
              "description": "List of the periods of time the user had the view in foreground (focused in the browser)",
```

ThibautGeriz: For the time being, it's only for browser but it should be usable as well from a mobile point of view.
At least, that's the idea.

schemas/view-schema.json
```diff
@@ -194,38 +194,26 @@
...
               "items": {
                 "type": "object",
-                "description": "Properties of the focused times of the view",
+                "description": "Properties of the foreground period of the view",
                 "required": ["duration", "start"],
                 "properties": {
                   "start": {
                     "type": "integer",
-                    "description": "Timestamp when the user gain focus of the document",
+                    "description": "Relative time in ns when in foreground from the view start",
```
vlad-mh:
```suggestion
                    "description": "Timestamp in ns relative to the view start when the event is in the foreground",
```
maybe?

schemas/view-schema.json
```diff
@@ -194,38 +194,26 @@
...
                     "minimum": 0,
                     "readOnly": true
                   },
                   "duration": {
                     "type": "integer",
-                    "description": "duration in ns when the user gain focus of the document",
+                    "description": "Duration in ns when in foreground",
```
vlad-mh:
```suggestion
                    "description": "Duration in ns when the event is in foreground",
```

schemas/view-schema.json
ThibautGeriz: I ended up looking at the definition of a timestamp and it's correct but feels weird to have a timestamp that's not starting at the same time as the UNIX one.
m2c

xgouchet (approved):

schemas/view-schema.json
bcaudan: What about having something similar to https://github.com/DataDog/rum-events-format/blob/d0bf4aaa7a95f8720cc9cefcf3d6f723049017bd/schemas/resource-schema.json#L89?
with view start and foreground start here

samples/view.json
```diff
@@ -41,7 +34,12 @@
     },
     "resource": {
       "count": 9
-    }
+    },
+    "in_foreground": true,
```
bcaudan: I am a bit confused about where we stand on collecting `in_foreground` for views.
Do we collect them then?

schemas/view-schema.json
```diff
@@ -194,38 +194,26 @@
...
-            "focused_times": {
+            },
+            "in_foreground_times": {
               "type": "array",
-              "description": "List of the time the user had the document focused for browser",
+              "description": "List of the period the user had the view in foreground (focused in the browser)",
               "items": {
                 "type": "object",
-                "description": "Properties of the focused times of the view",
+                "description": "Properties of the foreground period of the view",
```
bcaudan, line 203: it could be nice to use consistent wordings between the field name and the description.
- `in_foreground_times` -> a foreground time
- `in_foreground_periods` -> a foreground period
wdyt?

schemas/view-schema.json
bcaudan: Maybe we could reinforce the concept by having something like `Duration in ns of the view foreground time`

samples/view.json
ThibautGeriz: Oups I did not explicitly addressed that 😢 
I will remove them from the view to remove any possible confusion and only keep it in actions & errors.

louisj-dd (approved):

vlad-mh (approved):

schemas/_common-schema.json
```diff
@@ -81,19 +81,11 @@
...
-      "readOnly": true
-    },
-    "focus": {
-      "type": "object",
-      "description": "Focus properties",
-      "required": ["start_focused"],
-      "properties": {
-        "start_focused": {
+        },
+        "in_foreground": {
```
bcaudan: If we don't plan to have this field on all events, I'd be in favor of having it only on the corresponding events schema.

schemas/view-schema.json
```diff
@@ -194,38 +194,26 @@
...
                 "required": ["duration", "start"],
                 "properties": {
                   "start": {
                     "type": "integer",
-                    "description": "Timestamp when the user gain focus of the document",
+                    "description": "Duration in ns between start of the view and start of foreground",
```
bcaudan:
```suggestion
                    "description": "Duration in ns between start of the view and start of foreground period",
```

schemas/view-schema.json
```diff
@@ -194,38 +194,26 @@
...
                     "minimum": 0,
                     "readOnly": true
                   },
                   "duration": {
                     "type": "integer",
-                    "description": "duration in ns when the user gain focus of the document",
+                    "description": "Duration in ns when the event is in foreground",
```
bcaudan:
```suggestion
                    "description": "Duration in ns of the view foreground period",
```

bcaudan (approved):

PR #33, opened 2021-05-05
[RUMF-904] add the "context" field

BenoitZugmeyer:


xgouchet (approved):

schemas/_common-schema.json
```diff
@@ -182,6 +182,10 @@
         }
       },
       "readOnly": true
+    },
+    "context": {
+      "type": "object",
+      "description": "User provided context"
```
xgouchet: You should add the follow to allow anyone to add anything in there (this will be helpful for mobile side validation)
```
    "additionalProperties": {
        "type": "object"
    }
```

BenoitZugmeyer: This is already the case. FMU, if I add your suggestion, it will only accept properties that have objects as value. The validation fails: 

```
Validating samples/view.json ----------------------- ❌
Traceback (most recent call last):
  File "validate.py", line 19, in <module>
    jsonschema.validate(candidate, schema, resolver=resolver)
  File "/usr/local/lib/python2.7/site-packages/jsonschema/validators.py", line 934, in validate
    raise error
jsonschema.exceptions.ValidationError: u'bar' is not of type u'object'

Failed validating u'type' in schema[0][u'allOf'][0][u'properties'][u'context'][u'additionalProperties']:
    {u'type': u'object'}

On instance[u'context'][u'foo']:
    u'bar'
```

bcaudan (approved):
LGTM

benoitmeriaux (approved):

PR #34, opened 2021-05-17
[RUMF-906] add id on rum events

amortemousque:
### Motivation
Add id to all rum events

### Changes
Add id to errors, resources, long tasks and views
Id on actions is now required

### DIFF
--- a/schemas/action-schema.json
+++ b/schemas/action-schema.json
@@ -24,7 +24,8 @@
           "type": "object",
           "description": "Action properties",
           "required": [
-            "type"
+            "type",
+            "id"
           ],
           "properties": {
             "type": {
--- a/schemas/error-schema.json
+++ b/schemas/error-schema.json
@@ -24,6 +24,7 @@
           "type": "object",
           "description": "Error properties",
           "required": [
+            "id",
             "message",
             "source"
           ],
--- a/schemas/error-schema.json
+++ b/schemas/error-schema.json
@@ -28,6 +28,12 @@
             "source"
           ],
           "properties": {
+            "id": {
+              "type": "string",
+              "description": "UUID of the error",
+              "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+              "readOnly": true
+            },
             "message": {
               "type": "string",
               "description": "Error message",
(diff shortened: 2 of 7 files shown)


schemas/action-schema.json
```diff
@@ -130,8 +131,16 @@
...
           "type": "object",
           "description": "View properties",
-          "required": [],
+          "required": [
+            "id"
+          ],
           "properties": {
+            "id": {
+              "type": "string",
+              "description": "UUID of the view",
```
bcaudan: view id should already be in the event from common schema, no?

schemas/error-schema.json
```diff
@@ -24,10 +24,17 @@
           "type": "object",
           "description": "Error properties",
           "required": [
+            "id",
```
bcaudan, line 27: Let's not put new fields as required to ensure backward compatibility

schemas/error-schema.json
```diff
@@ -58,11 +65,18 @@
...
                 "status_code",
                 "url"
               ],
               "properties": {
+                "id": {
+                  "type": "string",
+                  "description": "UUID of the resource",
+                  "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+                  "readOnly": true
+                },
```
bcaudan: as discussed together, since resource errors should be removed soon, it can not worth the trouble.
And the goal of the task is more to identify each events rather than add extra links between events.

schemas/error-schema.json
```diff
@@ -141,4 +163,4 @@
       }
     }
   ]
-}
+}
```
bcaudan: there is no lint check in this repository for now, it could worth check your IDE settings to ensure that you are not removing EOF new lines

bcaudan (approved):

benoitmeriaux:
Could we add the id, as `required` for all the events, and not just for the view ?

amortemousque:
> Could we add the id, as `required` for all the events, and not just for the view ?

Sure, I haven't set ids as required for backward compatibility, but if all teams are aligned, then yes.

xgouchet (approved):

benoitmeriaux (approved):

bcaudan:
> if all teams are aligned, then yes

@amortemousque was there a discussion on that?

PR #37, opened 2021-05-27
Add Mobile Vitals metrics attributes

xgouchet:
This PR adds metrics related to the Mobile Vitals attributes collected on Mobile devices

### DIFF
--- a/schemas/view-schema.json
+++ b/schemas/view-schema.json
@@ -219,6 +219,64 @@
                 "readOnly": true
               },
               "readOnly": true
+            },
+            "metric": {
+              "type": "object",
+              "description": "Metrics of the view",
+              "properties": {
+                "memory": {
+                  "type": "object",
+                  "description": "Metrics related to the virtual memory used by the view",
+                  "properties": {
+                    "average": {
+                      "type": "number",
+                      "description": "Average memory used during the view lifetime (in bytes)",
+                      "readOnly": true
+                    },
+                    "max": {
+                      "type": "number",
+                      "description": "Peak memory used during the view lifetime (in bytes)",
+                      "readOnly": true
+                    }
+                  },
+                  "readOnly": true
+                },
+                "cpu": {
+                  "type": "object",
+                  "description": "Metrics related to the CPU used by the view",
+                  "properties": {
+                    "ticks": {
+                      "type": "number",
+                      "description": "Total number of cpu ticks during the view’s lifetime",
+                      "readOnly": true
+                    },
+                    "ticks_per_second": {
+                      "type": "number",
+                      "description": "Average number of cpu ticks per second during the view’s lifetime",
+                      "readOnly": true
+                    }
+                  },
+                  "readOnly": true
+                },
+                "refresh_rate": {
+                  "type": "object",
+                  "description": "Metrics related to the device refresh rate while the view is active",
+                  "properties": {
+                    "average": {
+                      "type": "number",
+                      "description": "Average refresh rate during the view’s lifetime (in frames per second)",
+                      "readOnly": true
+                    },
+                    "min": {
+                      "type": "number",
+                      "description": "Minimum refresh rate during the view’s lifetime (in frames per second)",
+                      "readOnly": true
+                    }
+                  },
+                  "readOnly": true
+                }
+              },
+              "readOnly": true
             }
           },
           "readOnly": true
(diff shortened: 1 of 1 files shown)


schemas/view-schema.json
```diff
@@ -219,6 +219,64 @@
                 "readOnly": true
               },
               "readOnly": true
+            },
+            "metric": {
```
BenoitZugmeyer: Existing metrics are directly in `view.*`. While I agree that it could be nice to group metrics in `view.metrics.*`, we lose a bit of consistency. Not a blocker IMO, but I'm curious of other opinions.

xgouchet: Hmm, that's true. I guess it's my habit of trying to order things but if everyone would rather have those metrics up one level we can do that.

bcaudan: agree, I would even go a bit further and remove the other grouping level:

```
view.memory_average
view.cpu_ticks_per_second
```

schemas/view-schema.json
```diff
@@ -219,6 +219,64 @@
...
+                  },
+                  "readOnly": true
+                },
+                "cpu": {
+                  "type": "object",
+                  "description": "Metrics related to the CPU used by the view",
+                  "properties": {
+                    "ticks": {
+                      "type": "number",
+                      "description": "Total number of cpu ticks during the view’s lifetime",
```
bcaudan: `ticks_count` could be more consistent with other namings

schemas/view-schema.json
```diff
@@ -219,6 +219,64 @@
...
               },
               "readOnly": true
+            },
+            "metric": {
+              "type": "object",
+              "description": "Metrics of the view",
+              "properties": {
+                "memory": {
+                  "type": "object",
+                  "description": "Metrics related to the virtual memory used by the view",
```
ncreated: "Used" is misleading (we can't guess which of pre-allocated memory was used). I think "allocated" is what we really mean, no?

schemas/view-schema.json
xgouchet: Wouldn't that make things confusing to not have the grouping levels ?

schemas/view-schema.json
xgouchet: Maybe used by the app during the view's lifetime is more accurate ?

schemas/view-schema.json
xgouchet: Also wouldn't this be more consistent with other values (we are using `view.error.count` and not `view.error_count`)

bcaudan: since v2, we only have nested attributes for core concepts and try to inline everything else:
```
// group for error
view.error.count

// no group for country
geo.country_iso_code
```

in the UI, we could still have:
- dedicated facet groups
- dedicated section in view side panel

schemas/view-schema.json
```diff
@@ -219,6 +219,36 @@
...
+              "type": "number",
+              "description": "Total number of cpu ticks during the view’s lifetime",
+              "readOnly": true
+            },
+            "cpu_ticks_per_second": {
+              "type": "number",
+              "description": "Average number of cpu ticks per second during the view’s lifetime",
+              "readOnly": true
+            },
+            "refresh_rate_average": {
```
0xnm, line 243: not sure if this one will be useful, especially if view is opened for a long time, but will see.

from practical point of view the most of the freezes/lags happen in the first 3-5 seconds since view is opened, probably - when everything is created from the scratch and data is loaded.

same probably applies to CPU, but with less degree - because of the background work.

0xnm (approved):

bcaudan (approved):

mariusc83 (approved):

schemas/view-schema.json
xgouchet: Yes but on average the duration on a given View on mobile can be short. Most of the time if the view spans over a long time it will probably have a few lags (e.g. infinite scroll on posts in a social media app where you're staying on the same activity and load new content every once in a while)

BenoitZugmeyer (approved):

louisj-dd (approved):
do you need these added to view documents ? aggregated somehow in sessions ?

xgouchet:
> do you need these added to view documents ? aggregated somehow in sessions ?
@louisj-dd yes we'd need to find those in the View document. I'm not sure they'd need to be aggregated in the session for now. We can do that later if need be.

PR #38, opened 2021-06-07
Add error.handling

amortemousque:
Add error.handling to identify if an error properly handled in the source code or not.

[RFC](https://docs.google.com/document/d/1aZuwrGMiZmh76xudk896FBFczjKkQKsqbjLvJMqrwAU/edit#)

### DIFF
--- a/schemas/error-schema.json
+++ b/schemas/error-schema.json
@@ -60,10 +60,9 @@
               "description": "The type of the error",
               "readOnly": true
             },
-            "handling": {
-              "type": "string",
+            "is_handled": {
+              "type": "boolean",
               "description": "Whether the error has been handled manually in the source code or not",
-              "enum": ["caught", "uncaught"],
               "readOnly": true
             },
             "resource": {
--- a/schemas/error-schema.json
+++ b/schemas/error-schema.json
@@ -60,9 +60,10 @@
               "description": "The type of the error",
               "readOnly": true
             },
-            "is_handled": {
-              "type": "boolean",
+            "handling": {
+              "type": "string",
               "description": "Whether the error has been handled manually in the source code or not",
+              "enum": ["handled", "unhandled"],
               "readOnly": true
             },
             "resource": {
--- a/schemas/error-schema.json
+++ b/schemas/error-schema.json
@@ -60,6 +60,12 @@
               "description": "The type of the error",
               "readOnly": true
             },
+            "handling": {
+              "type": "string",
+              "description": "Whether the error has been handled manually in the source code or not",
+              "enum": ["caught", "uncaught"],
+              "readOnly": true
+            },
             "resource": {
               "type": "object",
               "description": "Resource properties of the error",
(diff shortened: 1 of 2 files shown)


louisj-dd (approved):

xgouchet (approved):

schemas/error-schema.json
```diff
@@ -60,6 +60,12 @@
               "description": "The type of the error",
               "readOnly": true
             },
+            "handling": {
+              "type": "string",
+              "description": "Whether the error has been handled manually in the source code or not",
+              "enum": ["caught", "uncaught"],
+              "readOnly": true
+            },
```
BenoitZugmeyer: I'd like to discuss the name/value of this property a bit more.

* I think "caught" relates to a `try/catch` usage. But:
  * should a `console.error` be marked as "caught", since it has no `catch` involved?
  * rejected promises are generally associated with `handled/unhandled` instead of `caught/uncaught`

* Since "caught" and "uncaught" values are very binary (anything can be either caught or uncaught), I don't think we can add more value to this enum in the future. So, is it useful to have an enum instead of a boolean property?

In the RFC, we proposed `is_handled: bool`, and I wonder why we chose to implement it differently. I think the initial proposal is better.

xgouchet: I totally agree on the fact this could be a boolean instead of an enum. Also the handled/unhandled does seem more relevant than caught/uncaught as it applies to more use case than just a try-catch situation.

amortemousque: I also agree on these points. To me `is_handled` was more straightforward and less confusing for the reasons already explained above. 
Also, the `handling: caught/uncaught` adds more terms to understand and in terms of semantics, I'm not really convinced of the relation between "handling" and "caught".

maximethedoggo: Hello 👋, those are valid points 😬

In this case, would it be relevant to have a facet `Error Handling: handled/unhandled`? I do not really like boolean facets and the fact that we should have a long statement in the facet name like `Error Is Handled`, also it forces us to promote a value (in this case handled errors are promoted)

If you all disagree with ^ I am fine to keep `Error Is Handled: true/false` 👍

Also, we are planning to better surface this info on the FE, both in ET and RUM -> https://docs.google.com/document/d/1P69TWGlszXoQu0hFXTXN2Cjxz6UsAFgJeKkXeN98rOA/edit?usp=sharing

Having a boolean facet does not really help 😑

bcaudan: Having a enum rather than a boolean would indeed be more explicit in the UI.

BenoitZugmeyer: We already have multiple boolean fields: `is_crash`, `is_active`, `in_foreground`, `has_replay`. Using `is_handled` would be more in line with what we already have. If it's an UI issue, we may imagine changing the UI?

bcaudan: for the other boolean fields, it seems that they are either displayed as an icon or not being that important in the UI.
IMO, a solution avoiding to display true/false would be great.
I don't mind not using a boolean in this case if it makes things clearer for customers.

maximethedoggo: +1 for @bcaudan 

The good middle ground would be `Error Handling: handled / unhandled` no 😬?

amortemousque: If that make more sense for the end user, it seems reasonable :)

PR #39, opened 2021-06-09
Passing some mobile vitals metrics fields to integer

Manu-Tran:
This PR changes the type of some view fields related to mobile vitals metrics.
Memory_average/max and cpu_ticks_count and cpu_ticks_per_second should be long integers instead of floating numbers, especially with the given unit.

### DIFF
--- a/schemas/view-schema.json
+++ b/schemas/view-schema.json
@@ -221,22 +221,22 @@
               "readOnly": true
             },
             "memory_average": {
-              "type": "number",
+              "type": "integer",
               "description": "Average memory used during the view lifetime (in bytes)",
               "readOnly": true
             },
             "memory_max": {
-              "type": "number",
+              "type": "integer",
               "description": "Peak memory used during the view lifetime (in bytes)",
               "readOnly": true
             },
             "cpu_ticks_count": {
-              "type": "number",
+              "type": "integer",
               "description": "Total number of cpu ticks during the view’s lifetime",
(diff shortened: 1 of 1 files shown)


schemas/view-schema.json
```diff
@@ -221,22 +221,22 @@
...
             },
             "cpu_ticks_count": {
-              "type": "number",
+              "type": "integer",
               "description": "Total number of cpu ticks during the view’s lifetime",
               "readOnly": true
             },
             "cpu_ticks_per_second": {
-              "type": "number",
+              "type": "integer",
```
buranmert, line 239: this is `total / time`, result can be a floating point.
example: `cpu_ticks = 5` and `time = 2`, `cpu_ticks_per_second = 2.5`

the same argument holds for `memory_average` above as well

Manu-Tran: Thanks for the clarification ! That was what I thought, but what is the order of magnitude of the values ? I feel that if cpu_ticks is around ~10e5 or 10e6 then I would not know if using a float here is relevant, same for the memory since it is in bytes

buranmert: yes, i agree on the irrelevancy of fraction of a byte/cpu tick 👍

Manu-Tran: Understood, we'll do that backend side, I misunderstood the goal of this repo 👍

PR #41, opened 2021-06-24
[RUMF-936] add _dd.session.plan: 1|2 in the common schema

BenoitZugmeyer:
Motivation: RUM SDKs will introduce configuration options to let the customer chose which plan they want to use. This field will be used by:
* @DataDog/rum-back  to filter events according to the wanted plan and bill accordingly
* @DataDog/rum-app  to adapt the UI according to the chosen plan

### DIFF
--- a/schemas/_common-schema.json
+++ b/schemas/_common-schema.json
@@ -167,6 +167,17 @@
           "const": 2,
           "description": "Version of the RUM event format",
           "readOnly": true
+        },
+        "session": {
+          "type": "object",
+          "description": "Session-related internal properties",
+          "properties": {
+            "plan": {
+              "type": "number",
+              "description": "Session plan: 1 is the 'lite' plan, 2 is the 'replay' plan",
+              "enum": [1, 2]
+            }
+          }
         }
       },
       "readOnly": true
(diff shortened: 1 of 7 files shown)


vlad-mh (approved):

xgouchet (approved):

schemas/_common-schema.json
```diff
@@ -167,6 +167,17 @@
           "const": 2,
           "description": "Version of the RUM event format",
           "readOnly": true
+        },
+        "session": {
+          "type": "object",
+          "description": "Session-related internal properties",
+          "properties": {
+            "plan": {
```
bcaudan, line 176: what about putting plan field required in session object?

bcaudan (approved):

schemas/_common-schema.json
louisj-dd: but... didn't we say that we *cannot* add required fields ? :suspect:

bcaudan: in existing objects yes, but in a new optional object, it should not be an issue, right?

BenoitZugmeyer: Since `session` is a new field (optional), I guess we are free to have `plan` as required?

louisj-dd: 👌 didn't see it was under `_dd`

louisj-dd (approved):

PR #42, opened 2021-07-30
RUMM-1524 add Mobile Vitals related attributes

xgouchet:


schemas/view-schema.json
```diff
@@ -125,6 +125,11 @@
               "description": "Whether the View corresponding to this event is considered active",
               "readOnly": true
             },
+            "is_slow_rendered": {
+              "type": "boolean",
+              "description": "Whether the View had a low average refresh rate",
```
buranmert, line 130: how do we define _low_? we can mention it here to avoid confusion

mariusc83 (approved):

schemas/view-schema.json
xgouchet: Low is defined in the related RFC (an average framerate < 55 fps). But like `frozen_frame` we don't need to have the exact number is this may be changed later.

schemas/view-schema.json
```diff
@@ -181,6 +186,20 @@
               },
               "readOnly": true
             },
+            "frozen_frame": {
+              "type": "object",
```
0xnm, line 190: do we plan to extend `frozen_frame` in future to have not only `count` field?

xgouchet: No, not necessarily. We're following the standard set by the `action`, `error`, `crash`, `resource` and `long_task` fields

0xnm (approved):

schemas/view-schema.json
bcaudan: FMU, frozen frames are long tasks, would it make sense to instead have 

`view.long_task.frozen_frames_count`

it can depend on how much we want to promote this concept

xgouchet: We could do that as well, but for crashes (which are a subtype of errors) we do have `view.error.count` and `view.crash.count`, and not `view.error.crash_count`. IMO, it is better to have a `view.frozen_frame.count` as it's consistent with existing attribtues.

bcaudan: ok, sounds good to me

bcaudan (approved):

mariusc83 (approved):

schemas/view-schema.json
benoitmeriaux: Is it android (or mobile ?) specific ?
If the threshold can change later, should we not, store the `average_framerate` instead and compute the boolean in the backend or even at the query time with `average_framerate < 55` ?

schemas/view-schema.json
benoitmeriaux: consistency is good, but it seems that using object with only `xxx.count` is not so useful, do we really need to keep adding new object each time ?

schemas/view-schema.json
xgouchet: It's mobile specific indeed, the definition of how we compute it might change and be linked to data that's not part of the `average_framerate`. It's easier to let the SDK compute it.

schemas/view-schema.json
xgouchet: I won't fight for it , but consistency makes it easier for our customer to find attributes. If we start changing the naming paradigm now, it might add confusions to our customers.

benoitmeriaux: ok, let's keep it :)

benoitmeriaux (approved):

PR #45, opened 2021-09-08
RUMM-1604 Add a native resource type

xgouchet:
Adds a `native` resource type. 

## Additional details

Most existing resource types are specific to browser technology (`xhr`, `fetch`, …). So far, mobile SDKs have been using `xhr` as a default but this doesn't make any sense for mobile devs. Using a `native` type allows us to display a value that actually make sense.

### DIFF
--- a/schemas/resource-schema.json
+++ b/schemas/resource-schema.json
@@ -38,7 +38,7 @@
             "type": {
               "type": "string",
               "description": "Resource type",
-              "enum": ["document", "xhr", "beacon", "fetch", "css", "js", "image", "font", "media", "other"],
+              "enum": ["document", "xhr", "beacon", "fetch", "css", "js", "image", "font", "media", "other", "native"],
               "readOnly": true
             },
             "method": {


bcaudan (approved):

schemas/resource-schema.json
```diff
@@ -38,7 +38,7 @@
             "type": {
               "type": "string",
               "description": "Resource type",
-              "enum": ["document", "xhr", "beacon", "fetch", "css", "js", "image", "font", "media", "other"],
+              "enum": ["document", "xhr", "beacon", "fetch", "css", "js", "image", "font", "media", "other", "native"],
```
ncreated, line 41: I suppose we prefer `native` over `mobile`, as for example `xhr` makes sense in React Native? I'm only wondering if this will scale well to Flutter and Xamarin (and other cross-platforms) as some of them use native networking - e.g. AFAIK Xamarin uses `URLSession` on iOS (@maxep could you confirm 🙏?) so sending `native` type for Xamarin requests but `xhr` for RN could be problematic, no?

mariusc83 (approved):

schemas/resource-schema.json
xgouchet: Au contraire, being able to use `xhr` for React Native will be a good thing as it will make sense for customer. Using `native` for native SDKs / Flutter / Xamarin will be much less confusing than seeing `xhr`. And indeed we prefer `native` over `mobile` as it's slightly more generic and can be applicable to more topics. We could start adding the actual technology behind it (e.g.: `okhttp`, `urlsession`, …)  but it would be too noisy and eventually be a mess to display in the UI as well.

ncreated: I was thinking about `ios`, `android`, `xamarin` etc., but indeed that would require more effort and carefurness in UI and perhaps on backend (e.g. for generating trace). Let's start with `native`, as it already covers all our existing needs 👍.

ncreated (approved):

0xnm (approved):

schemas/resource-schema.json
maxep: Yep, I confirm that xamarin uses `NSURLSession`. Other implementations are [available](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/http-stack) tho, but the "managed" one is not recommended (no TLS > 1).

I also think `native` makes more sense, it also covers multiplatform apps (like Mac Catalyst).

maxep (approved):

schemas/resource-schema.json
benoitmeriaux: 🆗 for us, just to be clear, native = request from the app like xhr or fetch?  => we'll have to add the support for the apm span we generate currently only for xhr/fetch?

benoitmeriaux (approved):

PR #49, opened 2021-12-02
[ci-visibility] Add `"ci_visibility"` Property and `"ci-test"` Session Type

juan-fernandez:
Related PR: https://github.com/DataDog/browser-sdk/pull/1192 

* Add `ci_visibility` property.
* Update session type enum to include `'ci-test'`

### DIFF
--- a/schemas/_common-schema.json
+++ b/schemas/_common-schema.json
@@ -180,6 +180,19 @@
       },
       "readOnly": true
     },
+    "ci_visibility": {
+      "type": "object",
+      "description": "CI Visibility properties",
+      "required": ["test_execution_id"],
+      "properties": {
+        "test_execution_id": {
+          "type": "string",
+          "description": "The identifier of the current CI Visibility test execution",
+          "readOnly": true
+        }
+      },
+      "readOnly": true
+    },
     "_dd": {
       "type": "object",
       "description": "Internal properties",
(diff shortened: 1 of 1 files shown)


schemas/_common-schema.json
```diff
@@ -180,6 +180,19 @@
       },
       "readOnly": true
     },
+    "ci_visibility": {
```
mariusc83, line 183: I wonder how this section (property) would be used from let's say mobile sdk perspective ? Is this meant to be used on all platforms ?

juan-fernandez: I hadn't thought about it, but yeah, I don't see why not, there is no technology leakage. We currently use the trace id for the cypress poc, which could be reused for others

bcaudan (approved):

mariusc83 (approved):

Manu-Tran (approved):

schemas/_common-schema.json
```diff
@@ -45,7 +45,7 @@
         "type": {
           "type": "string",
           "description": "Type of the session",
-          "enum": ["user", "synthetics"],
+          "enum": ["user", "synthetics", "ci-test"],
```
bcaudan, line 48: what about having the same wording for both session type and extra context key?
`ci_visibility` VS `ci-test`

PR #53, opened 2022-01-11
RUMM-1901 Add 'source' property in the common schema

mariusc83:
As a follow up on the https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2290156382/RFC+-+Display+the+event+types+in+a+Hybrid+application RFC we are going to add a new `source` property in the `_common-schema.json`:
```
"source": {
      "type": "string",
      "description": "The source of this event",
      "enum": ["android", "ios", "browser", "flutter", "react-native"],
      "readOnly": true
    }
```

### DIFF
--- a/schemas/_common-schema.json
+++ b/schemas/_common-schema.json
@@ -56,6 +56,12 @@
       },
       "readOnly": true
     },
+    "source": {
+      "type": "string",
+      "description": "The source of this event",
+      "enum": ["android", "ios", "browser", "flutter", "react-native"],
+      "readOnly": true
+    },
     "view": {
       "type": "object",
       "description": "View properties",
--- a/samples/action.json
+++ b/samples/action.json
@@ -8,6 +8,7 @@
     "id": "cacbf45c-3a05-48ce-b066-d76349460599",
     "type": "user"
   },
+  "source": "android",
   "view": {
     "id": "623d50fd-75cf-4025-97d2-e51ff94171f6",
     "referrer": "",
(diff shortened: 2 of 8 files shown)


schemas/_common-schema.json
```diff
@@ -4,7 +4,7 @@
   "title": "CommonProperties",
   "type": "object",
   "description": "Schema of common properties of RUM events",
-  "required": ["date", "application", "session", "view", "_dd"],
+  "required": ["date", "application", "session", "view", "_dd", "source"],
```
bcaudan: cf [guideline](https://github.com/DataDog/rum-events-format#guidelines):
> New fields must not be set as required

buranmert (approved):

bcaudan (approved):

schemas/_common-schema.json
```diff
@@ -56,6 +56,12 @@
       },
       "readOnly": true
     },
+    "source": {
+      "type": "string",
+      "description": "The source of this event",
+      "enum": ["android", "ios", "browser", "flutter", "react-native", "cordova"],
```
xgouchet: Do we already know what the `cordova`/ `ionic`/`phonegap` value should be? Let's only add the known ones for now and add any new one later when it's also supported by the backend

xgouchet (changes requested):

louisj-dd (approved):

xgouchet (approved):

PR #55, opened 2022-03-02
RUMF-1175 Add report error source

amortemousque:
# Motivation

With the collection of `intervention` and `CSP violation` web reports, a new error source is needed.

# Changes

Add report error source in error schema


bcaudan (approved):

louisj-dd (approved):
ok for us but might be worth discussing this with error tracking team as well. I think they use this field

xgouchet (approved):

PR #56, opened 2022-03-09
[RUM] Update telemetry format

RomanWuattier:


xgouchet (approved):

0xnm (approved):

schemas/telemetry/debug-schema.json
```diff
@@ -9,19 +9,60 @@
...
-          "description": "Level/severity of the log",
-          "const": "debug"
-        },
-        "message": {
-          "type": "string",
-          "description": "Body of the log"
+        "telemetry": {
+          "type": "object",
+          "description": "The telemetry information as it should be indexed",
+          "required": ["status", "message"],
```
louisj-dd, line 17: are we sure to want to make both of these required ? maybe at some point we will want to add a metric telemetry and not send any log message for example ?

samples/telemetry/debug.json
```diff
@@ -1,16 +1,11 @@
...
   "date": 1591284175342,
   "service": "sample-sdk",
   "version": "1.2.3",
-  "status": "debug",
-  "message": "Session inconsistencies detected",
-  "debug": {
-    "cookie": "foo",
-    "session": "bar"
-  },
   "application": {
```
BenoitZugmeyer, line 10: Those fields should not be duplicated within the event.

schemas/telemetry/error-schema.json
```diff
@@ -9,31 +9,72 @@
...
+              "properties": {
+                "stack": {
+                  "type": "string",
+                  "description": "The stack trace or the complementary information about the error"
+                },
+                "kind": {
+                  "type": "string",
+                  "description": "The error type or kind (or code in some cases)"
+                },
+                "application": {
```
BenoitZugmeyer: **❓ question:** ‏Are those three fields expected to be nested in `telemetry.error`? Shouldn't it be in `telemetry` directly, like for debug telemetry events?

samples/telemetry/debug.json
RomanWuattier: It depends on whether we want to index those fields. Only what's in the `telemetry` will be indexed. Keeping these fields on the root is interesting for the processing phase.
I see 2 options. Either the SDK duplicates those fields, which might be easy since you know which ones exist (?). Or, the backend copies the fields but has to check for existence.

schemas/telemetry/error-schema.json
RomanWuattier: My bad that's correct

schemas/telemetry/debug-schema.json
xgouchet: this is just for a debug telemetry event. A metric telemetry event will be defined in another file, and indeed the status/message won't be mandatory

louisj-dd: oh, makes sense. lgtm then!

samples/telemetry/debug.json
0xnm: In this case why `action` field is not duplicated in the `Telemetry` object (and in `Telemetry.Error` in case of error event)?

BenoitZugmeyer: This feels strange to me. This implies that some part of the code has only access to the `telemetry` field, and some other part won't have access to the `telemetry` field? Is there a reason for that? Why not use the whole event everywhere?

RomanWuattier: Discussed it with the team: I removed the duplication, and the fields will stay at the root.
Answering your question, the RUM intake will ingest this event. We would keep some consistency between schemas, thus the metadata around the `telemetry` field. However, the logs intake is less restrictive and we can index the `telemetry` field containing the logs event only.
I pushed another commit.

xgouchet (approved):

RomanWuattier:
FYI: I pushed a new commit to add the `source` of the event. The field is interesting for debugging and logging purposes. It should not cause any trouble since it is present in every event (?)

xgouchet (approved):

schemas/telemetry/_common-schema.json
```diff
@@ -4,20 +4,27 @@
...
+          "type": "integer",
+          "const": 2,
+          "description": "Version of the RUM event format",
+          "readOnly": true
         }
       }
     },
+    "type": {
+      "type": "string",
+      "description": "RUM event type. Should specify telemetry only.",
```
BenoitZugmeyer: **🥜 nitpick:** ‏
```suggestion
      "description": "Telemetry event type. Should specify telemetry only.",
```
Those schemas are about "Telemetry events", versus schemas in the `schemas/rum` folder where we talk about "RUM events"

schemas/telemetry/debug-schema.json
```diff
@@ -9,19 +9,23 @@
...
-          "type": "string",
-          "description": "Level/severity of the log",
-          "const": "debug"
-        },
-        "message": {
-          "type": "string",
-          "description": "Body of the log"
+        "telemetry": {
+          "type": "object",
+          "description": "The telemetry information as it should be indexed",
```
BenoitZugmeyer: **🥜 nitpick:** I feel that "as it should be indexed" is unnecessary, since other fields should be indexed as well

BenoitZugmeyer (approved):

samples/telemetry/debug.json
```diff
@@ -19,5 +15,16 @@
...
+  "action": {
+    "id": "ae3a5d82-cdd1-468d-9bc9-3aa9e54d953c"
+  },
+  "telemetry": {
+    "status": "debug",
+    "message": "Session inconsistencies detected",
+    "debug": {
+      "cookie": "foo",
+      "session": "bar"
+    }
```
0xnm: This is not declared in the schema. Is schema not up-to-date?

RomanWuattier: Indeed, that's a good question. I kept what was previously defined. I don't know what specifications do you expect? Should we remove it or keep it?

amortemousque: This data is just to let us know that all data in the telemetry object is forwarded. Maybe you can add comment to make it clearer.

amortemousque (approved):

samples/telemetry/debug.json
RomanWuattier: Done

mariusc83 (approved):

PR #57, opened 2022-03-15
[RUMF-1206] Adjust Rum Event Format for Frustration Signals

BenoitZugmeyer:
# Motivation

Adjust RUM events for Frustration Signals needs

# Changes

* add an `@action.frustration_type: Array<'rage' | 'dead' | 'error'>` field on action events
* change `@action.id: uuid` to `@action.id: uuid | uuid[]` for events that can have action parent

# TypeScript changes:

```diff
diff --git packages/rum-core/src/rumEvent.types.ts packages/rum-core/src/rumEvent.types.ts
index 01d8bde8..9da26bba 100644
--- packages/rum-core/src/rumEvent.types.ts
+++ packages/rum-core/src/rumEvent.types.ts
@@ -41,6 +41,10 @@ export type RumActionEvent = CommonProperties & {
       name: string
       [k: string]: unknown
     }
+    /**
+     * Action frustration types
+     */
+    readonly frustration_type?: ('rage' | 'dead' | 'error')[]
     /**
      * Properties of the errors of the action
      */
@@ -98,7 +102,8 @@ export type RumActionEvent = CommonProperties & {
 /**
  * Schema of all properties of an Error event
  */
-export type RumErrorEvent = CommonProperties & {
+export type RumErrorEvent = CommonProperties &
+  ActionChildProperties & {
     /**
      * RUM event type
      */
@@ -195,16 +200,6 @@ export type RumErrorEvent = CommonProperties & {
       }
       [k: string]: unknown
     }
-  /**
-   * Action properties
-   */
-  readonly action?: {
-    /**
-     * UUID of the action
-     */
-    readonly id: string
-    [k: string]: unknown
-  }
     /**
      * View properties
      */
@@ -220,7 +215,8 @@ export type RumErrorEvent = CommonProperties & {
 /**
  * Schema of all properties of a Long Task event
  */
-export type RumLongTaskEvent = CommonProperties & {
+export type RumLongTaskEvent = CommonProperties &
+  ActionChildProperties & {
     /**
      * RUM event type
      */
@@ -243,22 +239,13 @@ export type RumLongTaskEvent = CommonProperties & {
       readonly is_frozen_frame?: boolean
       [k: string]: unknown
     }
-  /**
-   * Action properties
-   */
-  readonly action?: {
-    /**
-     * UUID of the action
-     */
-    readonly id: string
-    [k: string]: unknown
-  }
     [k: string]: unknown
   }
 /**
  * Schema of all properties of a Resource event
  */
-export type RumResourceEvent = CommonProperties & {
+export type RumResourceEvent = CommonProperties &
+  ActionChildProperties & {
     /**
      * RUM event type
      */
@@ -424,16 +411,6 @@ export type RumResourceEvent = CommonProperties & {
       }
       [k: string]: unknown
     }
-  /**
-   * Action properties
-   */
-  readonly action?: {
-    /**
-     * UUID of the action
-     */
-    readonly id: string
-    [k: string]: unknown
-  }
     /**
      * Internal properties
      */
@@ -827,3 +804,19 @@ export interface CommonProperties {
   }
   [k: string]: unknown
 }
+/**
+ * Schema of all properties of events that can have parent actions
+ */
+export interface ActionChildProperties {
+  /**
+   * Action properties
+   */
+  readonly action?: {
+    /**
+     * UUID of the action
+     */
+    readonly id: string | string[]
+    [k: string]: unknown
+  }
+  [k: string]: unknown
+}
```

### DIFF
--- a/schemas/rum/_action-child-schema.json
+++ b/schemas/rum/_action-child-schema.json
@@ -13,10 +13,21 @@
       ],
       "properties": {
         "id": {
-          "type": "string",
           "description": "UUID of the action",
-          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
-          "readOnly": true
+          "readOnly": true,
+          "oneOf": [
+            {
+              "type": "string",
+              "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
+            },
+            {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
+              }
+            }
+          ]
         }
       },
       "readOnly": true
--- a/schemas/rum/_action-child-schema.json
+++ b/schemas/rum/_action-child-schema.json
@@ -1,0 +1,26 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "_action-child-schema.json",
+  "title": "ActionChildProperties",
+  "type": "object",
+  "description": "Schema of all properties of events that can have parent actions",
+  "properties": {
+    "action": {
+      "type": "object",
+      "description": "Action properties",
+      "required": [
+        "id"
+      ],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "UUID of the action",
+          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+          "readOnly": true
+        }
+      },
+      "readOnly": true
+    }
+  }
+}
+
--- a/samples/rum/action.json
+++ b/samples/rum/action.json
@@ -29,7 +29,8 @@
     },
     "resource": {
       "count": 1
-    }
+    },
+    "frustration_type": ["rage"]
   },
   "_dd": {
     "format_version": 2,
--- a/samples/rum/error.json
+++ b/samples/rum/error.json
@@ -29,7 +29,10 @@
     }
   },
   "action": {
-    "id": "ae3a5d82-cdd1-468d-9bc9-3aa9e54d953c"
+    "id": [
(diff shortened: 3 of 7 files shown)


xgouchet (approved):
LGTM, will the backend be resilient to support both action.id with type string (old schema) and action.id with type array?

maxep:
With our strongly typed languages in mobile, I wonder how we would translate this `@action.id` change 🤔 . We are not that flexible for property type, we would basically have to pick one of the 2 types or create 2 distinct action models. WDYT @xgouchet @ncreated ?

BenoitZugmeyer:
> LGTM, will the backend be resilient to support both action.id with type string (old schema) and action.id with type array?

Discussed with @louisj-dd , it should be fine

> We are not that flexible for property type

Within the SDK, I think it is fine to only send arrays. But we need to keep the old type for retrocompatibility.

schemas/rum/_action-child-schema.json
```diff
@@ -0,0 +1,37 @@
...
+              "items": {
+                "type": "string",
+                "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
+              }
+            }
+          ]
+        }
+      },
+      "readOnly": true
+    }
```
ncreated, line 34: I agree with @maxep - the `oneOf` used here will be laborious to generate code in strongly typed languages. In Swift, we'd have to extend our generator and represent this with `enum`, sort of:
```swift
let id: SingleOrMultipleUUIDs

enum SingleOrMultipleUUIDs {
   case single(UUID)
   case multiple([UUID])
}
```
This notion is not something we currently support as we're not using general-purpose 3rd party generator (instead, we had to made our own). 

Is there a possibility to represent `action.id` in another way, possibly using existing schema constructs (e.g. only array, where single element would mean a single `id`)? Otherwise, we can do the work on our side, but as it requires additional time, it's IMO viable to consider alternative (if there is any and it can be correct).

BenoitZugmeyer: We have to keep the previous typing for retrocompatibility, but in SDKs it is fine to always send an array.

ncreated: Oh, I see 👌. So single `id` is coming from previous version of the schema and array is a new requirement 😥. Seems indeed that we don't have an easy option for that // cc @maxep

ncreated (approved):

louisj-dd:
lgtm but we need a few changes on backend side to be able to deal with uuid[] action.ids.

amortemousque (approved):

PR #59, opened 2022-04-27
[RUMF-1214] adjust frustration fields

BenoitZugmeyer:
* introduce a new field `view.frustration.count` to count frustrations occuring during the lifetime of the view
* rename `action.frustration_type` to `action.frustration.type` to match the counter field pattern

TypeScript diff:
```diff
@@ -42,9 +42,15 @@ export type RumActionEvent = CommonProperties & {
       [k: string]: unknown
     }
     /**
-     * Action frustration types
+     * Action frustration properties
      */
-    readonly frustration_type?: ('rage' | 'dead' | 'error')[]
+    readonly frustration?: {
+      /**
+       * Action frustration types
+       */
+      readonly type: ('rage' | 'dead' | 'error')[]
+      [k: string]: unknown
+    }
     /**
      * Properties of the errors of the action
      */
@@ -569,6 +575,16 @@ export type RumViewEvent = CommonProperties & {
       readonly count: number
       [k: string]: unknown
     }
+    /**
+     * Properties of the frustrations of the view
+     */
+    readonly frustration?: {
+      /**
+       * Number of frustrations that occurred on the view
+       */
+      readonly count: number
+      [k: string]: unknown
+    }
     /**
      * List of the periods of time the user had the view in foreground (focused in the browser)
      */
```

### DIFF
--- a/schemas/rum/action-schema.json
+++ b/schemas/rum/action-schema.json
@@ -60,14 +60,24 @@
               },
               "readOnly": true
             },
-            "frustration_type": {
-              "type": "array",
-              "description": "Action frustration types",
-              "readOnly": true,
-              "items": {
-                "type": "string",
-                "enum": ["rage", "dead", "error"]
-              }
+            "frustration": {
+              "type": "object",
+              "description": "Action frustration properties",
+              "required": [
+                "type"
+              ],
+              "properties": {
+                "type": {
+                  "type": "array",
+                  "description": "Action frustration types",
+                  "readOnly": true,
+                  "items": {
+                    "type": "string",
+                    "enum": ["rage", "dead", "error"]
+                  }
+                }
+              },
+              "readOnly": true
             },
             "error": {
               "type": "object",
--- a/samples/rum/action.json
+++ b/samples/rum/action.json
@@ -30,7 +30,9 @@
     "resource": {
       "count": 1
     },
-    "frustration_type": ["rage"]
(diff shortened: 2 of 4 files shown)


0xnm (approved):

schemas/rum/action-schema.json
```diff
@@ -60,14 +60,24 @@
...
+            "frustration": {
+              "type": "object",
+              "description": "Action frustration properties",
+              "required": [
+                "type"
+              ],
+              "properties": {
+                "type": {
+                  "type": "array",
+                  "description": "Action frustration types",
```
bcaudan, line 72: **💭 thought:** ‏idk if it was previously considered but since it is an array, should not it be `frustration.types`?

BenoitZugmeyer: It was not discussed, but IMO the singular version is fine:
* the fact that it is an array won't be user-facing too much, as most of the time the frustation will be rendered in a custom way by the UI
* when doing a query the user can specify on type at a time (ex: `@action.frustration.type:error`). Having `@action.frustration.types:error` looks a bit weird to me.

bcaudan: ok, lgtm

bcaudan (approved):

xgouchet (approved):

Manu-Tran (approved):

PR #62, opened 2022-05-24
RUMM-2195: Add OS and device information properties to the RUM schema

0xnm:
This change adds OS and device information properties to the RUM schema to stop relying on the User Agent parsing for mobile devices, because UA parsing is not quite reliable as a result of the wide range of UA formats which can be found across different Mobile/Tablet/TV/etc. devices.

This change is mostly backfill of the backend logic which extracts the following from the parsed UA:

* `os.name`
* `os.version`
* `os.version_major`
* `device.type`
* `device.name`
* `device.model`
* `device.brand`

All the fields of `os` property are required in assumption that it is always possible to get those values in any runtime.

Only `type` field of `device` is required, because SDK running on the `desktop`, for example, probably may not have any info about `name`, `model`, `brand` (or they can be not very representative).

`device.type` is taken from what we have already in the dashboard, with the only difference that `Appliance` type is replaced by the `TV`, because we have now a stronger focus on the TV platforms and any other device from `appliance` type can be marked as `other`.

### DIFF
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -204,6 +204,70 @@
       },
       "readOnly": true
     },
+    "os": {
+      "type": "object",
+      "description": "Operating system properties",
+      "required": [
+        "name",
+        "version",
+        "version_major"
+      ],
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Operating system name, e.g. Android, iOS",
+          "readOnly": true
+        },
+        "version": {
+          "type": "string",
+          "description": "Full operating system version, e.g. 8.1.1",
+          "readOnly": true
+        },
+        "version_major": {
+          "type": "string",
+          "description": "Major operating system version, e.g. 8",
+          "readOnly": true
+        }
+      }
+    },
+    "device": {
+      "type": "object",
+      "description": "Device properties",
+      "required": [
+        "type"
+      ],
+      "properties": {
+        "type": {
+          "type": "string",
+          "description": "Device type info",
+          "enum": [
+            "mobile",
+            "desktop",
+            "tablet",
+            "tv",
+            "gaming_console",
+            "bot",
+            "other"
+          ],
+          "readOnly": true
+        },
+        "name": {
+          "type": "string",
+          "description": "Device marketing name, e.g. Xiaomi Redmi Note 8 Pro, Pixel 5, etc.",
+          "readOnly": true
+        },
+        "model": {
+          "type": "string",
+          "description": "Device SKU model, e.g. Samsung SM-988GN, etc. Quite often name and model can be the same.",
+          "readOnly": true
+        },
+        "brand": {
+          "type": "string",
+          "description": "Device marketing brand, e.g. Apple, OPPO, Xiaomi, etc.",
+          "readOnly": true
+        }
+      }
+    },
     "_dd": {
       "type": "object",
       "description": "Internal properties",
--- a/samples/rum/view.json
+++ b/samples/rum/view.json
@@ -57,5 +57,16 @@
   },
   "context": {
     "foo": "bar"
+  },
+  "device": {
+    "type": "tablet",
+    "name": "iPad",
+    "model": "iPad",
+    "brand": "Apple"
+  },
+  "os": {
+    "name": "iOS",
+    "version": "15.1.1",
+    "version_major": "15"
   }
 }


schemas/rum/_common-schema.json
```diff
@@ -204,6 +204,70 @@
...
+        "type": {
+          "type": "string",
+          "description": "Device type info",
+          "enum": [
+            "mobile",
+            "desktop",
+            "tablet",
+            "tv",
+            "gaming_console",
+            "bot",
```
xgouchet, line 249: Do we have an accurate way to detect those natively? Also on the naming, I might use `gaming_system`, but not sure about this.

schemas/rum/_common-schema.json
```diff
@@ -204,6 +204,70 @@
...
+          "readOnly": true
+        }
+      }
+    },
+    "device": {
+      "type": "object",
+      "description": "Device properties",
+      "required": [
+        "type"
+      ],
```
xgouchet, line 238: Unless we upgrade the format version, all new fields are optionals

schemas/rum/_common-schema.json
```diff
@@ -204,6 +204,70 @@
...
       "readOnly": true
     },
+    "os": {
+      "type": "object",
+      "description": "Operating system properties",
+      "required": [
+        "name",
+        "version",
+        "version_major"
+      ],
```
xgouchet, line 214: Unless we upgrade the format version, all new fields are optionals

xgouchet (changes requested):

schemas/rum/_common-schema.json
0xnm: `os` and `device` parent fields are optional, so this is `required` rule only for the new optional parent fields, shouldn't break anything

schemas/rum/_common-schema.json
0xnm: I guess `gaming_console` may be deducted by having `Xbox`, `Playstation`, etc. in the device name or brand.

Not sure about `Bot` usage (maybe synthetics or case of web crawler?), but we have such type on the backend, so I decided to keep it just in case.

BenoitZugmeyer (approved):

schemas/rum/_common-schema.json
```diff
@@ -204,6 +204,70 @@
...
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Operating system name, ex. Android, iOS",
```
ncreated: nit: I believe it's [not the right abbreviation](https://www.quora.com/Is-example-abbreviated-as-i-e-ex-or-e-g) but I might be wrong 🙃
```suggestion
          "description": "Operating system name, e.g. Android, iOS",
```

schemas/rum/_common-schema.json
ncreated: Shouldn't we only add them once we need it? I think for now we can only use `mobile`, `tablet` and `tv`.

ncreated (approved):

schemas/rum/_common-schema.json
0xnm: yes, you are right

schemas/rum/_common-schema.json
0xnm: I added them all just in case if Browser SDK start using these properties as well, because this is what I see for example for the rum-app:

image

But I'm ok with shrinking this list as well.

xgouchet: I agree with @0xnm , these are the values available in the backend and since the browser (also using this rum format) can run on those, we need to be compatible

xgouchet (approved):

schemas/rum/_common-schema.json
BenoitZugmeyer: I'm fine with being exhaustive from the start or add it as needed. FYI we don't plan to provide those properties via the Browser SDK.

louisj-dd: [this](https://cs.github.com/DataDog/logs-backend/blob/3c8cc8db3f0a67ea7b25974423239feaa9efa7d0/domains/event-platform/libs/parsing/src/main/java/com/fsmatic/shared/parse/grok/UserAgentClassifier.java?q=UserAgentClassifier#L5) is the definition of this enum backend side

louisj-dd: ~i think you should try to replicate the backend enum values as much as possible~ nevermind i think it's fine this way as it keeps schema syntax consistent

louisj-dd (approved):

PR #64, opened 2022-06-08
[RUMF-1288] Add viewport size

amortemousque:
Add the viewport size to all events.

### DIFF
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -174,17 +174,17 @@
       "properties": {
         "viewport": {
           "type": "object",
-          "description": "Layout viewport properties",
+          "description": "The viewport represents the rectangular area that is currently being viewed. Content outside the viewport is not visible onscreen until scrolled into view.",
           "required": ["width", "height"],
           "properties": {
             "width": {
               "type": "number",
-              "description": "Width of the layout viewport (in pixels)",
+              "description": "Width of the viewport (in pixels)",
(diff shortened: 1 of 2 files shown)


BenoitZugmeyer (approved):

schemas/rum/_common-schema.json
```diff
@@ -168,6 +168,30 @@
         }
       }
     },
+    "display": {
+      "type": "object",
+      "description": "Display properties",
+      "properties": {
+        "viewport": {
```
0xnm, line 175: Will this property be used in some logic on the backend side? On mobile we normally don't have a definition of the viewport.

amortemousque: It will be used to filter clickmaps by viewport size. @xgouchet mention that it could also be interesting for mobile. Do you have something close to this notion?

JoelPfaffDD (approved):
👍

schemas/rum/_common-schema.json
0xnm: after the discussion with @amortemousque it is clear that `viewport` is just a rectangular area defining the user visible content (some content may be outside this area in case of the scrollable content). While `viewport` is not the term commonly used on mobile, I think we can still use it (and keep the name as is).

amortemousque: I've updated the definition. Let me know what you think.

0xnm (approved):

PR #65, opened 2022-06-23
Add Shape and Text json schemas for Mobile Session Replay

mariusc83:
In this PR we are introducing the JSON schemas for the Mobile Session Replay feature. 
The related RFC based on which the schemas were defined can be found [here](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2471460887/RFC+-+Data+Format+for+Mobile+Session+Replay)

For now we only introduced the JSON schemas for the `ShapeWireframe` and `TextWireframe` but more will come later.


validate.sh
```diff
@@ -2,16 +2,32 @@
...
-  else
-    echo "✅"
-  fi
-done
+declare -a sample_schema_folders=(
+  "rum-schema"
+  "session-replay-schema"
+  )
+
+declare -a event_formats_for_samples=(
```
mariusc83, line 10: Could not find a way to use `Map<String,String>` in bash so had to stick with arrays :)

xgouchet:
A couple of question, why add a new level in the samples folder hierarchy for `rum-schemas` vs `session-replay-schemas`, when there are already folders for  `rum` vs `telemetry`? 
Also maybe we should add `@DataDog/rum-session-replay` team as a code owner here

mariusc83:
> A couple of question, why add a new level in the samples folder hierarchy for `rum-schemas` vs `session-replay-schemas`, when there are already folders for `rum` vs `telemetry`? Also maybe we should add `@DataDog/rum-session-replay` team as a code owner here

@xgouchet Because the `rum` and `telemetry` folders are using the RUM schema format for validation, `session-replay` instead is using a different schema so I needed a way to group those for being able to make it work with the `validation.sh` script.
Regarding the code owners part is good idea, I will add this now.

schemas/session-replay/shape-wireframe-schema.json
```diff
@@ -0,0 +1,22 @@
...
+  "title": "ShapeWireframe",
+  "type": "object",
+  "description": "Schema of all properties of a ShapeWireframe.",
+  "allOf": [
+    {
+      "$ref": "_common-shape-wireframe-schema.json"
+    },
+    {
+      "properties": {
+        "type": {
```
ncreated, line 16: The `"type"` attribute should be `required` in all concrete wireframes, so here and in `text-wireframe-schema.json`.

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,69 @@
...
+  "type": "object",
+  "description": "Schema of common properties for ShapeWireframe events type and all its sub - types.",
+  "allOf": [
+    {
+      "$ref": "_common-wireframe-schema.json"
+    },
+    {
+      "required": [
+        "opacity",
+        "cornerRadius"
```
ncreated: Why making these two `required`? IMO:
- both have obvious default value for the player (`opacity: 1`, `cornerRadius: 0`);
- in practise majority of wireframes we record will use these defaults.

Making it required will only bloat the payloads in most cases where the wireframe uses default values. By making it optional we will safe 2 attributes per WF in majority of records.

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,69 @@
...
+          "description": "The opacity of this wireframe. Takes values from 0 to 1.",
+          "default": 1,
+          "readOnly": true
+        },
+        "cornerRadius": {
+          "type": "number",
+          "description": "The corner radius of this wireframe.",
+          "default": 0,
+          "readOnly": true
+        },
```
ncreated: Don't we think both are an aspect of the `shapeStyle`? IMO it's reasonable to make it an attribute of `shapeStyle`.

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,69 @@
...
+        "shapeStyle": {
+          "required": [
+            "fillColor"
```
ncreated: Similar, IMO nothing in there should be `required` and all properties can be optional as we can find sensible defaults. In case of `fillColor`, the default can be "transparent".

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,69 @@
...
+          ],
+          "type": "object",
+          "description": "The style of this wireframe.",
+          "properties": {
+            "fillColor": {
+              "type": "string",
+              "description": "The filling color for this wireframe.",
```
ncreated: > "The filling color for this wireframe."

Here we miss an information on the color format. I guess we assume hex string, but it's worth being explicit.

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,69 @@
...
+          "type": "object",
+          "description": "The stroke properties of this wireframe.",
+          "required": [
+            "color",
+            "width"
+          ],
+          "properties": {
+            "color": {
+              "type": "string",
+              "description": "The stroke color as a String hexadecimal.",
```
ncreated: I wonder if we're going to use `#RRGGBB` or `#RRGGBBAA` or allow both? Whereas it can be less complication for the player, mind that we need a custom implementation in recorders (iOS doesn't operate with hex colors and it uses R, G, B directly). It's IMO worth being more explicit.

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,133 @@
...
+  "title": "TextWireframe",
+  "type": "object",
+  "description": "Schema of all properties of a TextWireframe.",
+  "allOf": [
+    {
+      "$ref": "_common-shape-wireframe-schema.json"
+    },
+    {
+      "required": [
+        "text"
```
ncreated: Don't we also want to make `textStyle` required? It will be hard for the player to find sensible defaults to all text properties (including font size).

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,133 @@
...
+          "readOnly": true
+        },
+        "text": {
+          "type": "string",
+          "description": "The text value of the wireframe.",
+          "readOnly": false
+        },
+        "textStyle": {
+          "type": "object",
+          "description": "The style of this text field.",
```
ncreated: /nit - it's not only reserved for text fields. It can be used for "labels", "text areas", "segmented control titles" and many other platform-specific items - so better to stay platform agnostic.
```suggestion
          "description": "The style of the text.",
```

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,133 @@
...
+            "family": {
+              "type": "string",
+              "description": "The font family.",
+              "readOnly": true
+            },
+            "font": {
+              "type": "integer",
+              "description": "The font size.",
+              "readOnly": true
+            },
```
ncreated: I believe it should be just `"size"` not `"font"`.

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,133 @@
...
+            "color": {
+              "type": "string",
+              "description": "The font color as a string hexadecimal.",
```
ncreated: Similar - we should be more specific on the color format (`#RRGGBB` or `#RRGGBBAA`).

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,133 @@
...
+            }
+          },
+          "readOnly": true
+        },
+        "textPosition": {
+          "type": "object",
+          "description": "The position of the text in the wireframe.",
+          "required": [
+            "alignment",
+            "padding"
```
ncreated: Similar - both seem to have good defaults: `TL` + `0, 0, 0, 0` which will cover majority of use cases. IMO no reason to send it for every text wireframe.

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,133 @@
...
+          ],
+          "properties": {
+            "padding": {
+              "type": "object",
+              "required": [
+                "top",
+                "bottom",
+                "left",
+                "right"
```
ncreated: Similar - each defaults to `0`, so we can avoid the noise if of sending it.

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,133 @@
...
+                  "readOnly": true,
+                  "default": 0
+                }
+              }
+            },
+            "alignment": {
+              "type": "object",
+              "required": [
+                "horizontal",
+                "vertical"
```
ncreated: Similar - as we have good defaults to `0`, we can make it optional.

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,133 @@
...
+        },
+        "textStyle": {
+          "type": "object",
+          "description": "The style of this text field.",
+          "required": [
+            "family",
+            "font",
+            "color"
+          ],
+          "properties": {
```
ncreated, line 38: There was a good idea (i.a. from @louiszawadzki) to also include font type property which will help with finding a backup font (e.g. "serif" or "sans-serif").

ncreated (changes requested):
Great work 💪, definitely a good start. I left smaller comments - mainly on optionality of fields vs their defaults. 

Overal, I wonder on two things: 
- Shouldn't we call this format (the folder) a `mobile-session-replay` instead of `session-replay`? Both mean something different.
- Should we merge it directly to the `master` branch? I think it's fair to assume that at the beginning there will be a lot of iterations and breaking changes so maybe it's worth using feature branch for a while. On the other hand, both formats are isolated, so I'm ok with any approach.

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: Well the thing is that I don't see the reason to provide defaults for them in that case. If they are optionals, their default is `null`.

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: Yeah, it might be so. We can add them there.

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: Like I mentioned above, in case we mark a field as optional that means that it will default to `null`. In that case we don't need special defaults for it. Would you rather go with optionals for these fields ?

schemas/session-replay/text-wireframe-schema.json
mariusc83: Good point, I would like to use #RRGGBBAA here so I will specify it in the description.

schemas/session-replay/text-wireframe-schema.json
mariusc83: > Great work 💪, definitely a good start. I left smaller comments - mainly on optionality of fields vs their defaults.
> 
> Overal, I wonder on two things:
> 
> * Shouldn't we call this format (the folder) a `mobile-session-replay` instead of `session-replay`? Both mean something different.
> * Should we merge it directly to the `master` branch? I think it's fair to assume that at the beginning there will be a lot of iterations and breaking changes so maybe it's worth using feature branch for a while. On the other hand, both formats are isolated, so I'm ok with any approach.

@ncreated 
1. Regarding the format name I would rather keep it `session-replay` as this might relate with the browser session replay later. @BenoitZugmeyer What do you think ?
2. Like you also said, they are separated so I would rather merge them on master but I am waiting for others opinion also.

schemas/session-replay/text-wireframe-schema.json
mariusc83: We could...was following the RFC here but it might be better to make it non - optional

ThibautGeriz (commented):
Did not looking into details but this is only the shapes you're introducing aka the mobile FS content `data.wireframes`, not the records themselves. Do you plan to keep it that way?

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: On the other hand I get what you're saying, we can have it optional and add the `default` property just for visibility. This way at the player level they will know to which value to default if they can't find this option.

schemas/session-replay/_common-wireframe-schema.json
```diff
@@ -0,0 +1,42 @@
...
+    "x": {
+      "type": "integer",
+      "description": "The position on X axis of the UI element in absolute coordinates.",
+      "readOnly": true
+    },
+    "y": {
+      "type": "integer",
+      "description": "The position on Y axis of the UI element in absolute coordinates.",
+      "readOnly": true
+    },
```
ThibautGeriz: since we're talking about a shape, we should add a comment to which part of the shape we're talking about, top-left?

schemas/session-replay/_common-shape-wireframe-schema.json
ncreated: > Well the thing is that I don't see the reason to provide defaults for them in that case. If they are optionals, their default is null.

I don't mean providing defaults in JSON schema. I mean **implementing** defaults in the player. // cc @DataDog/rum-session-replay 

I think we mix optionality with nullability here, but these are two separate concepts. By including `opacity` as `required` parameter it will be generated as:
```swift
let opacity: Double
```
on our side. No option to set `null` (Kotlin) or `nil` (Swift).

If we don't list `opacity` in `required`, it will be optional:
```swift
let opacity: Double?
```
hence we will be able to pass `nil` making the field absent in the JSON or (in Swift) `Optional<Double>.none` making it use `null` in the JSON.


Ref.: [Required Properties](https://json-schema.org/understanding-json-schema/reference/object.html#required-properties) on json-schema.org.

schemas/session-replay/_common-wireframe-schema.json
mariusc83: @ThibautGeriz the idea is that I wanted to start small, I will add the records in another PR once this one is merged. Of course the records and segments will be added here.

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: Ok, that's clear now. We might go to provide a "default" in the JSON schema only for visibility but I am not sure. @ThibautGeriz what do you think, how do you prefer to have the defaults specified for each optional property ? Should  this be added in the schema also ?

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: I'd rather go with the #RRGGBBAA which covers both.

schemas/session-replay/_common-shape-wireframe-schema.json
ncreated: 👍 alternatively we could guide player's implementation by explaining each default in the description.

schemas/session-replay/shape-wireframe-schema.json
mariusc83: well, it is a constant so there's no need to mark it as required but will add it there for visibility.

schemas/session-replay/_common-shape-wireframe-schema.json
ThibautGeriz: I think it's fine to have default in make the property optional. It's even better if we can match the browser default. For example for opacity, it's no opacity

schemas/session-replay/_common-wireframe-schema.json
```diff
@@ -0,0 +1,42 @@
...
+    "y": {
+      "type": "integer",
+      "description": "The position on Y axis of the UI element in absolute coordinates. The anchor point will always be the top-left corner of the wireframe.",
+      "readOnly": true
+    },
+    "width": {
+      "type": "integer",
+      "description": "The width in pixels of the UI element normalised based on the device DPI.",
```
ThibautGeriz: > normalised based on the device DPI.

**question:** What does that mean?

mariusc83: Good point, I am going to detail more in the description.

schemas/session-replay/text-wireframe-schema.json
acorretti: > Shouldn't we call this format (the folder) a mobile-session-replay instead of session-replay? Both mean something different.

I think the directory can be called `session-replay` but then the `session-replay-events-schema.json` should reference mobile, e.g. `session-replay-events-mobile.json`. This makes room for browser ones to be added later. Wireframe definitions can keep their current names, there's little chance for collision with browser.

acorretti (changes requested):

.github/CODEOWNERS
```diff
@@ -1 +1,5 @@
 *   @Datadog/rum-back @Datadog/rum-browser @Datadog/rum-mobile
+
+# Session Replay
+/samples/session-replay-schema/ @DataDog/rum-session-replay
+/schemas/session-replay/ @DataDog/rum-session-replay
```
ThibautGeriz, line 5: **praise:** nice, thank you

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,58 @@
...
+          },
+          "opacity": {
+            "type": "number",
+            "description": "The opacity of this wireframe. Takes values from 0 to 1, default value should be 1.",
+            "readOnly": true
+          },
+          "cornerRadius": {
+            "type": "number",
+            "description": "The corner radius of this wireframe. The default value should be 0.",
+            "readOnly": true
```
ThibautGeriz: **question:** what's the unit of that: px? percentage? In the browser, we also can configure different radius for each corner, it's not possible in mobile?
for reference here is the browser configuration: https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,58 @@
...
+          },
+          "readOnly": true
+        },
+        "stroke": {
+          "type": "object",
+          "description": "The stroke properties of this wireframe.",
```
ThibautGeriz: **question:** stroke is the border? if so can you add that in the description so we have both name to help us do the mapping later on in the browser

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,58 @@
...
+          ],
+          "properties": {
+            "color": {
+              "type": "string",
+              "description": "The stroke color as a String hexadecimal. Follows the #RRGGBBAA color format.",
+              "readOnly": true
+            },
+            "width": {
+              "type": "integer",
+              "description": "The width of the stroke. The default value should be 1.",
```
ThibautGeriz: **thought:** not so sure we should have a default here, if there is a border, we should specify it's width. Well for browser it does not make sense but do mobile have mostly border of 1px? 🤔

schemas/session-replay/_common-wireframe-schema.json
```diff
@@ -0,0 +1,42 @@
...
+    "width": {
+      "type": "integer",
+      "description": "The width in pixels of the UI element normalised based on the device pixels per inch density (DPI). Example: if a device has a DPI = 2 the width of all UI elements will be divided to 2 to get a normalised width.",
+      "readOnly": true
+    },
+    "height": {
+      "type": "integer",
+      "description": "The height in pixels of the UI element normalised based on the device pixels per inch density (DPI). Example: if a device has a DPI = 2 the height of all UI elements will be divided to 2 to get a normalised height.",
+      "readOnly": true
+    }
```
ThibautGeriz, line 39: **praise:** thank you for clarifying
In reality we don't really care, we only care about the ratio, we'll just divide one by the other

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,121 @@
...
+            },
+            "size": {
+              "type": "integer",
+              "description": "The font size.",
+              "readOnly": true
+            },
+            "type": {
+              "type": "string",
+              "description": "The font type (serif, sans-serif, etc.).",
+              "readOnly": true
```
ThibautGeriz: **question:** shouldn't that be a enum?

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,121 @@
...
+            },
+            "color": {
+              "type": "string",
+              "description": "The font color as a string hexadecimal. Follows the #RRGGBBAA color format.",
```
ThibautGeriz, line 63: **question:** there is no regex for string in this framework?
**suggestion:** the `AA` part should/can be optional

ThibautGeriz (commented):
Looks really cool

samples/session-replay-mobile-schema/session-replay/shape.json
```diff
@@ -0,0 +1,17 @@
+{
+    "id": 1,
+    "x": 10,
+    "y": 10,
+    "width": 20,
+    "height": 30,
+    "opacity": 1.0,
+    "cornerRadius": 0.0,
```
jagracey, line 8: Consider naming `borderRadius` to match the browser?

samples/session-replay-mobile-schema/session-replay/shape.json
```diff
@@ -0,0 +1,17 @@
...
+    "shapeStyle": {
```
jagracey, line 9: Consider naming `backgroundColor` to match the browser for HTML (not XML).

jagracey:
I left two comments about naming conventions that seem to roughly follow SVG. Perhaps we should be following HTML/CSS naming conventions since IMO, we would probably not be using SVG.

schemas/session-replay/_common-shape-wireframe-schema.json
ncreated: > Well for browser it does not make sense but do mobile have mostly border of 1px? 🤔

I wouldn't say so. IMO most commonly there are no borders/strokes and surely there is no (iOS) guideline we can refer to. That said, IMO we should rather make `width` a `required` attribute and don't reason about the default. This way the real default that player can implement is "no stroke".

samples/session-replay-mobile-schema/session-replay/shape.json
mariusc83: I would rather keep it mobile oriented and you can do the analogy on the player side. @ncreated what do you think ?

samples/session-replay-mobile-schema/session-replay/shape.json
mariusc83: same as above, @ncreated need your opinion here also.

schemas/session-replay/text-wireframe-schema.json
mariusc83: We don't need a regex actually, but we can make the alpha optional. @ncreated thoughts ?

schemas/session-replay/text-wireframe-schema.json
mariusc83: Good question, I identified the following 5 types:  serif, sans serif, script, monospaced, and display. I think these should be enough, what do you think ?

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: px, I will add more info into the description.

schemas/session-replay/text-wireframe-schema.json
ThibautGeriz: I think let's type the field as much as we can and use an enum. If we missed one, let's add it after.

samples/session-replay-mobile-schema/session-replay/shape.json
ThibautGeriz: I agree, let's keep it mobile oriented, just make sure to add the browser equivalent in the description for clarity

schemas/session-replay/text-wireframe-schema.json
ThibautGeriz: I would not worry to much about that because in the end, the entry point is going to be segment with both browser & mobile datastructure

schemas/session-replay/_common-shape-wireframe-schema.json
ThibautGeriz: Sorry I am not sure I got the take away here, i see two cases:
- no border/stroke there nothing is the default
- there is a border/stroke therefore we need to give a width

Did I got it right?

mariusc83: yes that's it.

schemas/session-replay/text-wireframe-schema.json
ThibautGeriz: To be a bit more precise, the way I see it once we documented everything, it would look something like that:
```
schema/session-replay
schema/session-replay/segment.json
schema/session-replay/record.json
schema/session-replay/meta-record.json
schema/session-replay/focus-record.json
schema/session-replay/viewport-record.json
schema/session-replay/mobile/fullsnapshot-record.json
schema/session-replay/mobile/shape-wireframe.json
schema/session-replay/mobile/text-wireframe.json
schema/session-replay/browser/fullsnapshot-record.json
schema/session-replay/browser/serialized-node.json
```
What do you think?
(this is out of scope of the current PR)

mariusc83: Exactly what I was thinking also. If we all agree I can refactor to this folders architecture in the next PR where I am going to add the segment, records schemas.

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,58 @@
...
+    },
+    {
+      "properties": {
+        "shapeStyle": {
+          "type": "object",
+          "description": "The style of this wireframe.",
+          "properties": {
+            "fillColor": {
+              "type": "string",
+              "description": "The filling color for this wireframe as a String hexadecimal. Follows the #RRGGBBAA color format, default value should be #FFFFFF00.",
```
0xnm: just curious why `rgba` over `argb` (the latter is used on Android)?

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -0,0 +1,58 @@
...
+              "readOnly": true
+            }
+          },
+          "opacity": {
+            "type": "number",
+            "description": "The opacity of this wireframe. Takes values from 0 to 1, default value should be 1.",
```
0xnm, line 25: In theory the opacity is defined by the alpha component of the `shapeStyle#fillColor`, so we are kind of duplicating info. Or is it meant to be applied on top what if defined in the `shapeStyle#fillColor`? Say we have 0.5 here and 0.5 in the `shapeStyle#fillColor`, resulting alpha will be 0.25 or 0.5?

If it is a duplication, then it is ok for me, because a) it is not a lot of data; b) has a better clarity. Alternatively we can just keep rgb and alpha in the different fields. If it is an additional alpha, then I think we should add a notice to the description.

schemas/session-replay/_common-wireframe-schema.json
```diff
@@ -0,0 +1,42 @@
...
+  "title": "CommonWireframe",
+  "type": "object",
+  "description": "Schema of common properties for Wireframe events type.",
+  "required": [
+    "id",
+    "x",
+    "y",
+    "width",
+    "height",
+    "type"
```
0xnm: it is missing in the `properties` below. Should it be in the `required` list here? Seems it comes only with `ShapeWireframe`.

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -0,0 +1,128 @@
...
+          ],
+          "properties": {
+            "family": {
+              "type": "string",
+              "description": "The font family.",
+              "readOnly": true
+            },
+            "size": {
+              "type": "integer",
+              "description": "The font size.",
```
0xnm: I think we should specify a unit here. Is it supposed to be in points?

0xnm (commented):
Nicely done! I've left some comments, questions, but overall lgtm!

validate.sh
BenoitZugmeyer: You could move this logic to `validate.py` and enjoy a more evolved language :)

schemas/session-replay/text-wireframe-schema.json
BenoitZugmeyer: I like what @ThibautGeriz is suggesting. We'll consider adding browser-related schemas in the future.

ThibautGeriz: We'll have to it if we want to use the typescript type generated by the this repo.

BenoitZugmeyer: @ThibautGeriz can't you use both the Browser SDK types and the types from this repo at first?

schemas/session-replay/text-wireframe-schema.json
ncreated: Both (optional and non-optional `AA`) works for me. At the end, we will be exporting the value from r, g, b, a `floats` and constructing the hex manually, so I don't think we will have any preference.

samples/session-replay-mobile-schema/session-replay/shape.json
ncreated: Thing is, `shapeStyle` bundles more than background color: stroke width, stroke color, fill color, corner radius, opacity, ... . It could make sense to call `fillColor` a `backgroundColor`, but IMO not the `shapeStyle`.

schemas/session-replay/_common-shape-wireframe-schema.json
ncreated: > Or is it meant to be applied on top what if defined in the shapeStyle#fillColor?

Yes 👍.

My original thought behind this was to separate wireframe's `opacity` from fill / stroke opacity (embedded in `#RRGGBBAA`). Otherwise, when exporting wireframes we'd need to perform blending on the SDK side to result with proper values in hex colors.

samples/session-replay-mobile-schema/session-replay/shape.json
jagracey: Similarly there are a few more styles I was thinking of, like fill and stroke being SVG based as well.
Anyway no big deal at this point.

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: I think the `RRGGBBAA` is more popular

schemas/session-replay/_common-wireframe-schema.json
mariusc83: not it is a left over from a prev iteration ;)

schemas/session-replay/text-wireframe-schema.json
mariusc83: We can keep it as it is but and have in mind that we can also skip the alpha if needed.

validate.sh
mariusc83: Yes, did not want to do this in this PR, maybe later in a dedicated PR.

schemas/session-replay/text-wireframe-schema.json
ThibautGeriz: @BenoitZugmeyer Kinda but it's gonna be awkward, we'll start with that. As I target I wish we could avoid. The segment we fetch we'll have to cast it `MobileSegment | BrowserSegment` with both having the same definition for the focus record for example

ThibautGeriz (approved):

acorretti (approved):

ncreated (approved):
Well done 💪

PR #66, opened 2022-06-29
RUMM-2319 Add schemas for Segments and Records

mariusc83:
In this PR we are adding the schemas for a `Segment` and `Record` data types following the [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2471460887/RFC+-+Data+Format+for+Mobile+Session+Replay).

Please note:

- the examples will be added later in a different PR once we will generate the models locally in our SDKs
- for the `IncrementalData` types we just introduced the interface schema -> `incremental-data-schema.json` and the actual subtypes schemas will be added in a different PR.


schemas/session-replay/metadata-record-schema.json
```diff
@@ -0,0 +1,47 @@
...
+                        "width": {
+                            "type": "integer",
+                            "description": "The width of the screen in pixels normalised based on the device pixels per inch density (DPI). Example: if a device has a DPI = 2 the normalized width will be the current width divided by 2.",
+                            "readOnly": true
+                        },
+                        "height": {
+                            "type": "integer",
+                            "description": "The height of the screen in pixels normalised based on the device pixels per inch density (DPI). Example: if a device has a DPI = 2 the normalized height will be the current height divided by 2.",
+                            "readOnly": true
+                        }
```
ThibautGeriz, line 42: **issue:** here we need to know if it's mobile or browser. We should be able to know the following answer:
- should we show the keyboard/mobile frame (disregarding if it's a mobile app or a mobile website)
- should we using the web or mobile renderer (aka is it DOM or Wireframe)

**suggestion:** add the following fields:
- `source` to be `browser`, `ios`, `android` (if missing it would be `browser` for retro compatibility) (matching the RUM attribute `source`)
- `deviceType` or `device_type` to be either `Desktop`, `Mobile`, `Tablet`, `Bot`, `Appliance`, `Other`  (if missing it would be `Desktop` for retro compatibility) (matching the RUM attribute `device.type`)


cc @BenoitZugmeyer

schemas/session-replay/incremental-data-schema.json
```diff
@@ -0,0 +1,10 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "incremental-data-schema.json",
+    "title": "IncremenetalData",
+    "type": "object",
+    "description": "Schema of a Session Replay IncrementalData type.",
+    "oneOf": [
+        {}
```
acorretti: question: is this incomplete?

schemas/session-replay/incremental-data-schema.json
```diff
@@ -0,0 +1,10 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "incremental-data-schema.json",
+    "title": "IncremenetalData",
```
acorretti: nit: Typo `IncrementalData`

schemas/session-replay/incremental-data-schema.json
mariusc83: Yes...this will come in another PR

schemas/session-replay/view-end-record-schema.json
```diff
@@ -0,0 +1,26 @@
...
+    "type": "object",
+    "description": "Schema of a Record which signifies that view lifecycle ended.",
+    "allOf": [
+        {
+            "$ref": "_common-record-schema.json"
+        },
+        {
+            "required": [
+                "type",
+                "data"
```
acorretti: question: is `data` defined?

schemas/session-replay/_common-record-schema.json
```diff
@@ -0,0 +1,17 @@
...
+    "type": "object",
+    "description": "Schema of common properties for a Record event type.",
+    "required": [
+        "timestamp"
+    ],
+    "properties": {
+        "timestamp": {
+            "type": "integer",
+            "description": "Defines the UTC time in milliseconds when this Record was performed.",
+            "readOnly": true
```
ncreated, line 14: I assume it should be computed relatively to 00:00:00 UTC on 1 January 1970, but this info is missing in the `description`. Also, we don't expect it to get negative, so we can define `minimum: 0` as we do in [RUM events](https://github.com/DataDog/rum-events-format/blob/b01e4821e5dd59e8fab780eae68329e7b4deabf1/schemas/rum/_common-schema.json#L12). This way we will be able to safely generate `Int64` type for holding this property.

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -16,6 +16,7 @@
           "properties": {
             "backgroundColor": {
               "type": "string",
+              "pattern": "^[#][a-f0-9]{2}[a-f0-9]{2}[a-f0-9]{2}([a-f0-9]{2})?",
```
ThibautGeriz: **suggestion:** simplify and add `$` at the end: `^#[a-f0-9]{6}([a-f0-9]{2})?$`

https://regex101.com/r/61Jo1w/1

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -42,6 +43,7 @@
           "properties": {
             "color": {
               "type": "string",
+              "pattern": "^[#][a-f0-9]{2}[a-f0-9]{2}[a-f0-9]{2}([a-f0-9]{2})?",
```
ThibautGeriz: **suggestion:** simplify and add `$` at the end: `^#[a-f0-9]{6}([a-f0-9]{2})?$`

https://regex101.com/r/61Jo1w/1

schemas/session-replay/_common-shape-wireframe-schema.json
```diff
@@ -16,6 +16,7 @@
           "properties": {
             "backgroundColor": {
               "type": "string",
+              "pattern": "^[#][a-f0-9]{2}[a-f0-9]{2}[a-f0-9]{2}([a-f0-9]{2})?",
               "description": "The background color for this wireframe as a String hexadecimal. Follows the #RRGGBBAA color format, default value should be #FFFFFF00.",
```
ncreated: Description says `#RRGGBBAA`, but the pattern allows for `#RRGGBB` too - it's inconsistent. Also, the pattern is missing end-of-string `$`. I think we can simplify it to:
```
"pattern": "^#[a-f0-9]{8}$"
```
no?

schemas/session-replay/focus-record-schema.json
```diff
@@ -0,0 +1,30 @@
...
+                "type": {
+                    "type": "integer",
+                    "description": "The type of this Record.",
+                    "const": 3,
+                    "readOnly": true
+                },
+                "data": {
+                    "type": "boolean",
+                    "description": "Whether this screen has a focus or not. For now it will always be true for mobile.",
+                    "const": true
```
ThibautGeriz: **issues:** this is not true for browser
**suggestion:** remove `"const": true` ~and add in the description that it's always true for mobile session replay~

schemas/session-replay/full-snapshot-record-schema.json
```diff
@@ -0,0 +1,33 @@
+{
```
ThibautGeriz: Move to `schemas/session-replay/mobile/full-snapshot-record-schema.json`

schemas/session-replay/incremental-data-schema.json
```diff
@@ -0,0 +1,10 @@
+{
```
ThibautGeriz: Move to `schemas/session-replay/mobile/incremental-data-schema.json`

schemas/session-replay/focus-record-schema.json
```diff
@@ -0,0 +1,30 @@
...
+                }
```
ncreated: The web format uses `has_focus` attribute instead of `data: Bool`:
```typescript
data: {
   has_focus: boolean // always `true`
}
```
Do we want to change it?

schemas/session-replay/full-snapshot-record-schema.json
```diff
@@ -0,0 +1,33 @@
...
+                    "readOnly": true
+                },
+                "data": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "wireframe-schema.json"
+                    },
+                    "description": "The Wireframes contained by this Record.",
+                    "readOnly": true
+                }
```
ncreated: IMO it's safer to define `data.wireframes: Wireframe[]` (not `data: Wireframe[]`), like proposed in the RFC:
```typescript
data: {
   wireframes: ConcreteWireframe[]
}
```
I mean "safer" as it leaves room for adding other properties to `data` in backward-compatible fashion (without altering the entire type).

schemas/session-replay/segment-schema.json
```diff
@@ -0,0 +1,115 @@
...
+                "records_count": {
+                    "type": "integer",
+                    "description": "The number of records in this Segment.",
+                    "readOnly": true
+                },
+                "creation_reason": {
+                    "type": "string",
+                    "description": "The reason this Segment was created. For now for mobile we only have one possible value for this and will always be the default value.",
+                    "enum": [
+                        "init"
```
ThibautGeriz: **issue:** browser has more here
```ts
export type CreationReason =
  | 'init'
  | 'segment_duration_limit'
  | 'segment_bytes_limit'
  | 'view_change'
  | 'before_unload'
  | 'visibility_hidden'
```

https://github.com/DataDog/browser-sdk/blob/main/packages/rum/src/types/segment.ts#L22-L28

**questions:**
- should we add them?
- can you think of specific ones for mobile based on your HTTP flush strategy?

cc @BenoitZugmeyer

schemas/session-replay/incremental-snapshot-record-schema.json
```diff
@@ -0,0 +1,34 @@
...
+    "type": "object",
+    "description": "Schema of a Record type which contains mutations of a screen.",
+    "allOf": [
+        {
+            "$ref": "_common-record-schema.json"
+        },
+        {
+            "required": [
+                "type",
+                "data"
```
ncreated: It's a bit inconsistent if we require `data` in incremental record, but not in some other (focus + full snapshot). Is this intentional?

schemas/session-replay/incremental-snapshot-record-schema.json
```diff
@@ -0,0 +1,34 @@
...
+            ],
+            "properties": {
+                "type": {
+                    "type": "integer",
+                    "description": "The type of this Record.",
+                    "const": 3,
```
ncreated: `3` is already used in full snapshot ⚠️

schemas/session-replay/segment-schema.json
```diff
@@ -0,0 +1,115 @@
...
+                    ],
+                    "default": "init",
+                    "readOnly": true
+                },
+                "index_in_view": {
+                    "type": "integer",
+                    "description": "The index of this Segment in the segments list that was recorded for this view ID.",
+                    "readOnly": true
+                },
```
ThibautGeriz: **suggestion(non-blocking):** add in the description that it starts at 0

schemas/session-replay/text-wireframe-schema.json
```diff
@@ -60,6 +60,7 @@
             },
             "color": {
               "type": "string",
+              "pattern": "^[#][a-f0-9]{2}[a-f0-9]{2}[a-f0-9]{2}([a-f0-9]{2})?",
```
ThibautGeriz: **suggestion:** simplify and add `$` at the end: `^#[a-f0-9]{6}([a-f0-9]{2})?$`

https://regex101.com/r/61Jo1w/1

schemas/session-replay/view-end-record-schema.json
ThibautGeriz: No it's not: https://github.com/DataDog/browser-sdk/blob/main/packages/rum/src/types/record.ts#L59-L62
it should be removed from here

ThibautGeriz (commented):
Looking good, I think it's start to organise a bit more our schema, I suggest we use something like that:
```
schema/session-replay
schema/session-replay/segment.json
schema/session-replay/record.json
schema/session-replay/meta-record.json
schema/session-replay/focus-record.json
schema/session-replay/viewport-record.json
schema/session-replay/mobile/fullsnapshot-record.json
schema/session-replay/mobile/shape-wireframe.json
schema/session-replay/mobile/text-wireframe.json
schema/session-replay/browser/fullsnapshot-record.json
schema/session-replay/browser/serialized-node.json
```

schemas/session-replay/segment-schema.json
```diff
@@ -0,0 +1,115 @@
...
+                "start": {
+                    "type": "integer",
+                    "description": "The start UTC timestamp in milliseconds corresponding to the first record in the Segment data.",
+                    "readOnly": true
+                },
+                "end": {
+                    "type": "integer",
+                    "description": "The end UTC timestamp in milliseconds corresponding to the last record in the Segment data.",
+                    "readOnly": true
+                },
```
ncreated: If we assume "from epoch" (00:00:00.000 01.01.1970) in both, then it's IMO worth adding in description.

ncreated (changes requested):
Nice 👌. I left few minor feedbacks with a blocking one on `type` value conflict.

schemas/session-replay/incremental-snapshot-record-schema.json
mariusc83: nope...I missed that...should be required everywhere and `focus` doesn't have any data.

schemas/session-replay/metadata-record-schema.json
mariusc83: Should this only be in the `MetadataRecord` ?

schemas/session-replay/_common-shape-wireframe-schema.json
mariusc83: actually we decided that we will have the alpha optional.I will modify the description to state that.

schemas/session-replay/focus-record-schema.json
mariusc83: Nope, was just a mistake.

schemas/session-replay/metadata-record-schema.json
mquentin: I agree with Thibaut, having the same source field as the rum events is needed. This context is required to initialize the renderer and we don't want to wait for an initial RUM track query to get this info.

Adding it to any record might be a good thing but, in the renderer part we will use the metadata-record to switch into mobile or browser mode.

schemas/session-replay/segment-schema.json
mariusc83: reason why I did not want to add them for now is that they are quite browser specific. I think when the time comes to add browser schemas here we could have a `_common-segment-schema` and then 2 `mobile-segment-schema` and `browser-segment-schema` which inherit from it. What do you think ? Do you think it makes sense to add them now already ?

schemas/session-replay/metadata-record-schema.json
mariusc83: What about the `device_type` ? @ThibautGeriz @mquentin why do you think this one is needed ?

schemas/session-replay/segment-schema.json
amortemousque: It's true that some creation reasons are specific to browser. If the segment schema only defer by this list, I don't mind having one list with both browser and mobile reasons.

schemas/session-replay/segment-schema.json
```diff
@@ -0,0 +1,115 @@
...
+                "end": {
+                    "type": "integer",
+                    "description": "The end UTC timestamp in milliseconds corresponding to the last record in the Segment data. Each timestamp is computed as the UTC interval since 00:00:00.000 01.01.1970.",
+                    "readOnly": true
+                },
+                "has_full_snapshot_record": {
+                    "type": "boolean",
+                    "description": "Whether this Segment contains a fullsnapshot record or not.",
```
0xnm: minor typo:

```suggestion
                    "description": "Whether this Segment contains a full snapshot record or not.",
```

schemas/session-replay/segment-schema.json
```diff
@@ -0,0 +1,115 @@
...
+                    "readOnly": true
+                },
+                "records_count": {
+                    "type": "integer",
+                    "description": "The number of records in this Segment.",
```
0xnm, line 100: maybe a dumb question: why do we need `records_count` field when we have `records` array field and can simply know the count? Is it because records for the same view may arrive in different requests?

0xnm (approved):

schemas/session-replay/segment-schema.json
mariusc83: Good question, is part of the `Segment` web event contract. Maybe @ThibautGeriz, @acorretti can give us more info on this matter.

acorretti: The current schema doesn't make this clear, but these properties are part of what's called the `SegmentMetadata` that is not only present in each segment, but also as individual attachments in the _multipart POST_ request to the intake endpoint.

It's like this so the intake can still know some details about the segment without having to decompress them.

Maybe it's a good idea to define the _segment metadata_ separately, then have _segment-schema_ extend it?

schemas/session-replay/metadata-record-schema.json
ThibautGeriz: > Should this only be in the MetadataRecord ?

yes, this is the initial record that kick everything else in motion, once we know it's all good

> What about the device_type ? why do you think this one is needed ?

It's not needed at the moment but the moment we want to add something related to the platform like keyboard or some phone frame, it's going to be needed, I think it's better to be thorough on the data-structure

schemas/session-replay/segment-schema.json
ThibautGeriz: > 2 mobile-segment-schema and browser-segment-schema which inherit from it. What do you think ? Do you think it makes sense to add them now already ?

I don't think we want two different segment definitions, they are very close. Well time will tell, you can leave that as is

schemas/session-replay/full-snapshot-record-schema.json
ThibautGeriz: Indeed better, it's closer to the browser definition:
```ts
export interface FullSnapshotRecord {
  type: typeof RecordType.FullSnapshot
  timestamp: TimeStamp
  data: {
    node: SerializedNodeWithId
    initialOffset: {
      top: number
      left: number
    }
  }
}
```

schemas/session-replay/segment-schema.json
ThibautGeriz: > Good question, is part of the Segment web event contract. can give us more info on this matter.

As Alfonso said, they is the segment that's compressed and is described here but we also send the metadata in the HTTP payload as multi part field in order to index them in the event platform without having to decompress the payload stored on S3. So the `records` field in only in S3 while `records_count` is also in the event platform
 
> Maybe it's a good idea to define the segment metadata separately, then have segment-schema extend it?

I don't think it's worth

schemas/session-replay/wireframe-schema.json
```diff
@@ -0,0 +1,15 @@
+{
```
ThibautGeriz: Move to `schemas/session-replay/mobile/wireframe-schema.json`

schemas/session-replay-mobile-events-schema.json
```diff
@@ -6,10 +6,10 @@
...
     {
-      "$ref": "session-replay/shape-wireframe-schema.json"
+      "$ref": "session-replay/mobile/shape-wireframe-schema.json"
     },
     {
-      "$ref": "session-replay/text-wireframe-schema.json"
+      "$ref": "session-replay/mobile/text-wireframe-schema.json"
     }
   ]
 }
```
ThibautGeriz: **question:** Shouldn't that thing reference the segment schema now?

ThibautGeriz (commented):
Looking good

schemas/session-replay-mobile-events-schema.json
mariusc83: yeah I need to add here those refs also.

schemas/session-replay-mobile-events-schema.json
```diff
@@ -6,10 +6,16 @@
   "description": "Schema of all properties of a mobile Session Replay event",
```
ThibautGeriz, line 6: **nitpic:** you can remove the `Mobile` since it's now records & segments that are shared across platform

schemas/session-replay/metadata-record-schema.json
```diff
@@ -0,0 +1,60 @@
...
+                    "description": "The source of this record",
+                    "enum": [
+                        "android",
+                        "ios",
+                        "browser",
+                        "flutter",
+                        "react-native"
+                    ],
+                    "readOnly": true
+                },
```
ThibautGeriz: **suggestion:** move this to data and make it mandatory

ThibautGeriz: **question:** what about `device_type`?

schemas/session-replay/mobile/_common-shape-wireframe-schema.json
```diff
@@ -16,7 +16,8 @@
           "properties": {
             "backgroundColor": {
               "type": "string",
-              "description": "The background color for this wireframe as a String hexadecimal. Follows the #RRGGBBAA color format, default value should be #FFFFFF00.",
+              "pattern": "^[#][a-f0-9]{6}([a-f0-9]{2})?$",
```
ThibautGeriz: **question:** why keep the bracket around the `#` It's not necessary, right?

schemas/session-replay/mobile/text-wireframe-schema.json
```diff
@@ -60,7 +60,8 @@
             },
             "color": {
               "type": "string",
-              "description": "The font color as a string hexadecimal. Follows the #RRGGBBAA color format.",
+              "pattern": "^[#][a-f0-9]{2}[a-f0-9]{2}[a-f0-9]{2}([a-f0-9]{2})?",
```
ThibautGeriz: **suggestion:** simplify regex like the others

schemas/session-replay-mobile-events-schema.json
```diff
@@ -6,10 +6,16 @@
...
     },
     {
-      "$ref": "session-replay/text-wireframe-schema.json"
+      "$ref": "session-replay/mobile/text-wireframe-schema.json"
+    },
+    {
+      "$ref": "session-replay/segment-schema.json"
+    },
+    {
+      "$ref": "session-replay/record-schema.json"
```
ThibautGeriz: **question:** why add the subtypes and not just the segment that includes all of them?

schemas/session-replay-mobile-events-schema.json
mariusc83: My idea was that once you will add your own schemas here we will have a `session-replay-browser-events-schema.json` also

schemas/session-replay/metadata-record-schema.json
mariusc83: Let's leave this for later.

schemas/session-replay/mobile/_common-shape-wireframe-schema.json
mariusc83: Well....is the way you define a color as a hexa string. I would rather keep it.

schemas/session-replay-mobile-events-schema.json
mariusc83: good point

schemas/session-replay/metadata-record-schema.json
mariusc83: Actually on the first point...I am not sure this should be at the `data` level. The `source` is a different topic than the actual `data`. What do you think ?

ThibautGeriz: > Actually on the first point...I am not sure this should be at the `data` level. The `source` is a different topic than the actual data. What do you think ?

No the source is the data of the meta record. the only field outside of data is the timestamp, let's keep it that way

> Let's leave this for later.

😢

schemas/session-replay/mobile/_common-shape-wireframe-schema.json
ThibautGeriz: Yeah but in the regex, you can change `^[#][a-f0-9]{6}([a-f0-9]{2})?$` to `^#[a-f0-9]{6}([a-f0-9]{2})?$`

schemas/session-replay/metadata-record-schema.json
mquentin: Ok so all things considered I think it's nice to have the source field mandatory for all the record types, not only the `MetadataRecord`.

On our side it could allow us to keep the players agnostic and leave the `if mobile then` to the sandbox and if we know the source for each record it will simply the implementation

schemas/session-replay/metadata-record-schema.json
mquentin: > Should this only be in the MetadataRecord ?

cf. https://github.com/DataDog/rum-events-format/pull/66#discussion_r913742244
I think it would be useful to have the source for all records

schemas/session-replay/metadata-record-schema.json
mariusc83: that means this will be a part of the `_common-record-schema.json`. Do we all agree here ? @ThibautGeriz @mquentin @ncreated @acorretti ?

ThibautGeriz: I would rather leave them on the meta record honestly, a small state is better than duplicating the field everywhere. We would only need it on the fullsnapshot & incremental. So if we want to avoid state, we should only add it to those. But I am not super fan of this.

I am curious to have your opinion @acorretti & @BenoitZugmeyer

schemas/session-replay/segment-schema.json
BenoitZugmeyer: The creation reason was reported to help debugging unknown issues. I don't think it is ever used in production, so I don't think it matters too much.

IMO a simple `string` without enumerating the values would be fine. But I don't mind being explicit and enumerate values here. More values can be added when we port the Browser types, no need to list them all now?

About having different segment for Browser and Mobile: if possible, let's avoid that :)

schemas/session-replay/metadata-record-schema.json
BenoitZugmeyer: > have the source field mandatory for all the record types

This would be against [the guidelines](https://github.com/DataDog/rum-events-format/#guidelines): since older versions of the Browser SDK would not send the `source` field, you cannot assume that it will be always defined. Beside that, we should avoid adding 10+ bytes to every records.

What about moving this information to the Segment instead?
* no need to maintain a state from the last Metadata record
* we would be able to know the source even if there is no Metadata record in a segment (might be useful when debugging)
* it could be indexed (not sure if useful but still)

ThibautGeriz: actually, Benoit shared his opinion here https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2471460887/RFC+-+Data+Format+for+Mobile+Session+Replay?focusedCommentId=2515730503

> What would be the use-case? I think that it would be more convenient to have it as a segment meta property than in a record, because to won’t need to find the Meta record inside a Segment to access the information. Also, Segments don’t always have Meta records.
> Also I have the secret ambition to remove the Meta record one day. URL is useless, and for the viewport size we have the visualviewport record.


I see three options. Am I missing something?


### 1. Add it in the segment metadata (pick)

Add it in the segment metadata. Once received, the fetcher will add to the `RecordWithViewId` data-structure and the renderer will have have it with all records

**Pros:**
- have the info on all the segments (we don't need it right now)
- no duplication at the sdk level
- no need to keep a state in the HTML renderer
- Benoit can possibly complete his secret dream


**Cons:**
- need to add logic in the player itself (no big deal)
- internal message a little big bigger (very small consequence)
- the initialisation state is split between the meta record & segment itself
  - what about if we want to add the device_type if we were to add it, where should it go?

### 2. Add it in the meta record

The meta record create the sandbox and keep the source as a react state

**Pros:**
- we already have all the initial state in there to create a sandbox like the URL (used for the base tag) or the initial dimension
- no duplication at the sdk level
- simple

**Cons:**
- we need to add a new state HTML renderer
- Benoit cannot complete his secret dream 😿 


### 3. Add it in all records

**Pros:**
- have the info on all the record (we don't need it right now)
- no need to keep a state in the HTML renderer
- Benoit can possibly complete his secret dream


**Cons:**
- duplication at the sdk level
- internal message a little big bigger (very small consequence)
- the initialisation state is split between the meta record & inside all records
  - what about if we want to add the device_type, where should it go?

I would choose 1 or 2

ThibautGeriz: > This would be against [the guidelines](https://github.com/DataDog/rum-events-format/#guidelines): since older versions of the Browser SDK would not send the source field, you cannot assume that it will be always defined. Beside that, we should avoid adding 10+ bytes to every records.

it's fine, if it's not here, we'll say it's browser

mariusc83: so we add it in the `MetadataRecord#data#source` ?

mquentin: Indeed I get it,
either we go with the first or second option. I like the first one but we already maintain a state built on the meta record so even the second option works for me.

anyway, in both case with will provide the `isMobile` context to the sandbox iframe using the message bridge.

ThibautGeriz (approved):
LGTM

ncreated (approved):
LG 👌

PR #67, opened 2022-07-05
RUMM-2320 JSON schemas for IncrementalData types

mariusc83:
This PR adds the remaining JSON schemas for the `IncrementalData` types which will go in the `IncrementalSnapshotRecord#data`.

### DIFF
--- a/schemas/session-replay/mobile/mutation-data-schema.json
+++ b/schemas/session-replay/mobile/mutation-data-schema.json
@@ -1,0 +1,68 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "mutation-data-schema.json",
+    "title": "MutationData",
+    "type": "object",
+    "description": "Schema of a MutationData.",
+    "allOf": [
+        {
+            "properties": {
+                "source": {
+                    "type": "integer",
+                    "const": "1",
+                    "description": "The source of this type of incremental data.",
+                    "readOnly": true
+                },
+                "adds": {
+                    "type": "array",
+                    "readOnly": true,
+                    "items": {
+                        "type": "object",
+                        "required": [
+                            "wireframe"
+                        ],
+                        "properties": {
+                            "previousId": {
+                                "type": "integer",
+                                "description": "The previous wireframe id next/after which this new wireframe should be drawn/attached to.",
+                                "readOnly": true
+                            },
+                            "wireframe": {
+                                "$ref": "wireframe-schema.json",
+                                "description": "The new wireframe that needs to be added to the screen.",
+                                "readOnly": true
+                            }
+                        }
+                    },
+                    "description": "Contains the newly added wireframes."
+                },
+                "removes": {
+                    "type": "array",
+                    "readOnly": true,
+                    "items": {
+                        "type": "object",
+                        "required": [
+                            "id"
+                        ],
+                        "properties": {
+                            "id": {
+                                "type": "integer",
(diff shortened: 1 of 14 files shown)


mquentin (approved):
LGTM

did you changes the blocks you extracted in their own json files like `schemas/session-replay/mobile/text-style-schema.json` ? (to see if I need to re-generate the shape and text wireframes types)

mariusc83:
> LGTM
> 
> did you changes the blocks you extracted in their own json files like `schemas/session-replay/mobile/text-style-schema.json` ? (to see if I need to re-generate the shape and text wireframes types)

Nope, I just extracted them, nothing was changed.

schemas/session-replay/mobile/mutation-data-schema.json
```diff
@@ -0,0 +1,69 @@
...
+                    "description": "The source of this type of incremental data.",
+                    "readOnly": true
+                },
+                "adds": {
+                    "type": "array",
+                    "readOnly": true,
+                    "items": {
+                        "type": "object",
+                        "required": [
+                            "previousId",
```
ncreated, line 22: We need `previousId` to accept `null` value, hence it cannot be `required`. This is to enable placing the new `wireframe` at the beginning of the array.

Ref. in RFC: [Adding new wireframes](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2471460887/RFC+-+Data+Format+for+Mobile+Session+Replay)
> wireframe in draw order (or null if the new wireframe should be rendered at the very bottom).

schemas/session-replay/mobile/mutation-data-schema.json
```diff
@@ -0,0 +1,69 @@
...
+                            "wireframe"
+                        ],
+                        "properties": {
+                            "previousId": {
+                                "type": "integer",
+                                "description": "The previous wireframe id next/after which this new wireframe should be drawn/attached to.",
+                                "readOnly": true
+                            },
```
ncreated: I'm not sure this is correct. We need the `previousId` to support `null` value in order to place new `wireframe` at the beginning of the array.

Ref. in RFC: [Adding new wireframes](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2471460887/RFC+-+Data+Format+for+Mobile+Session+Replay)
> Besides the wireframe to add, it holds the ID of the previous wireframe in draw order (or null if the new wireframe should be rendered at the very bottom).

So, the allowed values for `previousId` should be: `Int | null`. Doesn't it mean we need to enhance the `type` definition for `previousId`? I'm not sure how to model this in [JSON schema](https://json-schema.org/understanding-json-schema/reference/object.html#id8). 

@mariusc83  @BenoitZugmeyer any thoughts?

BenoitZugmeyer: Good catch. Indeed, if you want to put the wireframe at the very beginning of the list, `previousId` should not be defined. IMO an optional property would be fine, no need to have `Int | null` explicitely.

ncreated: 👍 Let's make it not `required` then. In fact, in Swift-generated code both `Int | null | none` are expressed with optional, so no preference on our side.

mariusc83: Yup, good catch, we will make it not required.

ncreated (approved):
Looks good 👌!

PR #68, opened 2022-07-05
Add feature flags to telemetry event

amortemousque:
Since the browser SDK uses standard feature flags (cf: https://github.com/DataDog/web-ui/pull/60432) we can now sample them.
The goal is to help us investigate an experimental feature impact by comparing two data sets:
- One with the feature enabled
- Another with the feature disabled

To do so, we also need to add the enabled feature flag to the telemetry events to filter on them.

### DIFF
--- a/schemas/telemetry/_common-schema.json
+++ b/schemas/telemetry/_common-schema.json
@@ -92,6 +92,14 @@
           "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
         }
       }
+    },
+    "experimental_features":  {
+      "type": "array",
+      "description": "Enabled experimental features",
+      "items": {
+        "type": "string"
+      },
+      "readOnly": true
     }
   }
 }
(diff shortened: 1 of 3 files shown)


schemas/telemetry/_common-schema.json
```diff
@@ -92,6 +92,14 @@
           "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$"
         }
       }
+    },
+    "experimental_features":  {
```
bcaudan, line 96: **❓ question:** ‏Do @DataDog/rum-mobile have a similar concept?

RomanWuattier, line 96: Indexing this new value requires a minor update of the backend.
Speaking of indexing, do you want the list to be stored in the log's body as is?

RomanWuattier (approved):
👍 for rum-back

schemas/telemetry/_common-schema.json
xgouchet: We don't but this is something we could have eventually

xgouchet (approved):

bcaudan (approved):

schemas/telemetry/_common-schema.json
amortemousque: I'd say yes. Do you see any limitation with this format?

RomanWuattier: No

PR #69, opened 2022-07-11
Small fixes on the session-replay schemas

mariusc83:


schemas/session-replay/mobile/mutation-data-schema.json
```diff
@@ -9,15 +9,14 @@
...
                     "readOnly": true
                 },
                 "adds": {
                     "type": "array",
                     "readOnly": true,
                     "items": {
-                        "type": "object",
                         "required": [
                             "wireframe"
                         ],
```
ncreated, line 22: Wouldn't removing `"type": "object"` make the schema definition for array element invalid? AFAIK every schema (including this subschema) must have `type` definition, no? I couldn't find an answer in [JSON Schema `type` spec](https://json-schema.org/understanding-json-schema/reference/type.html).

ncreated (approved):
Looks good 👍, with one question on `type` removal.

schemas/session-replay/mobile/mutation-data-schema.json
mariusc83: we do not need a type there because we already have this schema so it will be matched by id, at least this is what I had in mind. I guess we will need web team input on this one. 
@ThibautGeriz @mquentin We need your input here. Basically the question is :
In the mutation data (wireframes) do you still need the type of the wireframe having in mind that this wireframe can be matched by `id` in your current screen ?

mariusc83: @ncreated It will not make it invalid, I just added a new PR where I am validating all the schemas with new samples. Basically the mutated wireframes inherit from a different `_common`

mariusc83: @ncreated sorry I was out of topic on that question. I thought it was referring to the `type` of the wireframe update 🤦 . I don't think it matters as long as it is a reference. The reference schema gives the type after.

mquentin: > In the mutation data (wireframes) do you still need the type of the wireframe having in mind that this wireframe can be matched by id in your current screen ?

I don't think we will store a map of `Id + wireframe` (that contains the Type), we will most likely store a map of `Id + htmlRepresentationOfTheWireframe`.

It means that, having the type of the wireframe in the mutation will allow us not to guess the type of the `htmlRepresentationOfTheWireframe`.

You can still find a workaround in our implementation tho but I confident that having the type will help us.

mquentin (approved):
LGTM

schemas/session-replay/mobile/mutation-data-schema.json
mariusc83: Great, is good that we discussed this. If you don't mind I am going to add this in the next PR (the one that is already opened where I am also adding the samples).

PR #70, opened 2022-07-12
RUMM-2321 Add Session Replay JSON samples

mariusc83:
This PR adds the required Session Replay JSON sample to be matched with the available schemas.

Please note that additional fixes were added to the current schemas based on the validation results.


schemas/session-replay/mobile/wireframe-update-mutation-schema.json
```diff
@@ -7,9 +7,6 @@
     "oneOf": [
         {
             "$ref": "text-wireframe-update-schema.json"
-        },
-        {
-            "$ref": "shape-wireframe-update-schema.json"
```
mariusc83: Damn...I will add it back 🤦

samples/session-replay-mobile-schema/session-replay/incremental-snapshot-record.json
```diff
@@ -0,0 +1,53 @@
+{
+    "timestamp": 1657620232337,
+    "type": 3,
+    "data": [
+        {
+            "source": 1,
```
mquentin: Ok so regarding that, the version I have from the browser sdk does not match:

```
export enum IncrementalSource {
    Mutation = 0,
    MouseMove = 1,
    MouseInteraction = 2,
    Scroll = 3,
    ViewportResize = 4,
    Input = 5,
    TouchMove = 6,
    MediaInteraction = 7,
    StyleSheetRule = 8
}
```

it should be source=1 isn't it ?

mariusc83: Hmm I think that's a different thing. The data on mobile defers from the one on web. See here our mutation data: 
https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2471460887/RFC+-+Data+Format+for+Mobile+Session+Replay

samples/session-replay-mobile-schema/session-replay/incremental-snapshot-record.json
```diff
@@ -0,0 +1,53 @@
+{
+    "timestamp": 1657620232337,
+    "type": 3,
+    "data": [
```
mquentin: Here I also have a question, in the browser SDK equivalent, it is not a list but a unique `IncrementalData` and inside the `MutationData` contains the list. Is it a required feature to have list here ? Can we receive a list of mutations at once that are not homogeneous like this example `[MutationData, MousemoveData, MouseInteractionData, ScrollData]` ?

mariusc83: Oh I think I see what you say...it might be a problem there...let me look into this. I will come back with a fix.

mquentin (approved):
LGTM

samples/session-replay-mobile-schema/session-replay/segment.json
```diff
@@ -0,0 +1,110 @@
...
+                "removes": [
+                    {
+                        "id": 3
+                    }
+                ],
+                "updates": [
+                    {
+                        "id": 2,
+                        "text": "New text",
+                        "type": "text",
```
ncreated: Schema shouldn't allow changing the `type` of a wireframe in incremental update. For example, if the already player accumulated a state for "text" wireframe and we change it to "shape", what should it do with `text` properties?

schemas/session-replay/mobile/incremental-snapshot-record-schema.json
```diff
@@ -21,10 +21,7 @@
                     "readOnly": true
                 },
                 "data": {
-                    "type": "array",
-                    "items": {
-                        "$ref": "incremental-data-schema.json"
-                    },
+                    "$ref": "incremental-data-schema.json",
                     "description": "The IncrementalData contained by this Record.",
                     "readOnly": true
```
ncreated, line 26: Won't `description` clash with what is defined in `$ref`'s `description`?

samples/session-replay-mobile-schema/session-replay/segment.json
mariusc83: this is needed for @mquentin on his end and also it's important for us as our generator doesn't know to create a deserializer yet if you have:
```
A(id:Int)
B(id:Int, size:Int)
```
We need that type const there.

schemas/session-replay/mobile/incremental-snapshot-record-schema.json
mariusc83: Yes, but there's a merging mechanism, the `@ref` one takes priority.

bcaudan (approved):
LGTM

PR #73, opened 2022-07-22
Add CPU architecture to RUM error events

fuzzybinary:
The CPU architecture will be needed by Flutter to determine which symbols file to use to decode Dart stack traces.

### DIFF
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -289,6 +289,11 @@
           "type": "string",
           "description": "Device marketing brand, e.g. Apple, OPPO, Xiaomi, etc.",
           "readOnly": true
+        },
+        "architecture": {
+          "type": "string",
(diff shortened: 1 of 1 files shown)


schemas/rum/error-schema.json
```diff
@@ -80,6 +80,11 @@
               "enum": ["android", "browser", "ios", "react-native", "flutter"],
               "readOnly": true
             },
+            "architecture": {
```
BenoitZugmeyer: **💬 suggestion:** ‏What about adding this as `device.architecture` attribute? https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_common-schema.json#L257

(It could be added to all events)

fuzzybinary: Yeah, that's a good call. I'll move it there.

BenoitZugmeyer (approved):

mariusc83 (approved):

PR #71, opened 2022-07-20
REPLAY-922: Add browser Session Replay schema definitions

acorretti:
## What is this PR doing?

Adds Browser Session Replay schemas, and unifies schema organisation together with Mobile Session Replay.
* Sets three entrypoints for Session Replay's `mobile`, `browser` and combined `mobile + browser` schemas:
```
(combined) schemas/session-replay-events-schema.json
(browser) schemas/session-replay-browser-events-schema.json
(mobile) schemas/session-replay-mobile-events-schema.json
```
* Mobile samples validation now distinguishes between `segment` validations, and `record` validations.
* Includes latest additions in regards to `FrustrationRecord` and `FrustrationTypes`.

### Best reviewed by commit
Upon merging, commits will be squashed.

## QA
These changes have been validated in the [Browser SDK](https://github.com/DataDog/browser-sdk) by generating the Typescript definitions (with [json-schema-to-typescript](https://github.com/bcaudan/json-schema-to-typescript)) and swapping the existing types by the generated ones.

Mobile SDK recorder team has validated the generated types both in iOS & Android.

PR #74, opened 2022-07-26
REPLAY-943: Simplify entrypoint files and save a top-level type alias

acorretti:
## What is this PR doing?

Out of PR #71 we ended up with two entrypoint Session Replay schema definitions: one for mobile, another for browser.
Due to how our generators worked, we needed each entrypoint to define a `type` as well as reference the respective `segment-schema`s.

The pitfall of that approach is that we end up with an unnecessary top-level type alias.
E.g. browser's:
```
export type BrowserSessionReplay = BrowserSegment
```
As we adjusted our generators to lift this requirement, we can also get rid of this type level alias.

## QA

Re-generated Browser SDK's & Mobile SDK's types, made sure the generated types were only getting rid of the top-level aliases.


mariusc83 (approved):

ncreated (approved):
LG 👌

BenoitZugmeyer (approved):

PR #77, opened 2022-07-27
RUMM-2374 Add support for OTA updates in error schema

louiszawadzki:
# Purpose of this PR

This PR adds support for OTA (over the air) updates for javascript cross-platform technologies such as React Native.

OTA Updates enable developers to update the javascript bundle of their app remotely, without having to submit their app to the stores:
image

When using OTA updates we need 2 attributes to match an error to its source maps:
- the native version (generated when the app was submitted to the stores)
- the bundle version (generated when the bundle is built)

The main OTA updates services for react native are [CodePush](https://docs.microsoft.com/en-us/appcenter/distribution/codepush/) and [Expo EAS Update](https://docs.expo.dev/eas-update/introduction/).

We have access to the bundle version in the app, so we can send it to the backend when there's a crash to unminify the error using the correct sourcemaps.
image

The idea is then to display this information as a facet when it is available in the UI to quickly see if a bug is fixed in a new update:

image

### DIFF
--- a/schemas/rum/error-schema.json
+++ b/schemas/rum/error-schema.json
@@ -131,6 +131,11 @@
                 }
               },
               "readOnly": true
+            },
+            "bundle_version": {
+              "type": "string",
+              "description": "Version of the javascript bundle used in the case of Over The Air updates",
+              "readOnly": true
             }
           },
           "readOnly": true


schemas/rum/error-schema.json
```diff
@@ -131,6 +131,11 @@
                 }
               },
               "readOnly": true
+            },
+            "bundle_version": {
```
nioufe, line 135: I think we need to use version for this. 
A lot of datadog systems use version to detect regressions and such. Introducing a new version "facet" could be problematic.

nioufe (changes requested):
Putting it as request changes so that we have a "product" discussion but if those discussion happened, happy to remove it

BenoitZugmeyer:
> When using OTA updates we need 2 attributes to match an error to its source maps:
> 
>   *  the native version (generated when the app was submitted to the stores)
>    * the bundle version (generated when the bundle is built)

Why do you need the native version? Is it a `Source Code Integration` product requirement?

If you need both versions, I suggest to colocate them: add `bundle_version` as a root attribute of all events [similarly to `version`](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_common-schema.json#L34)

schemas/rum/error-schema.json
louiszawadzki: This is a very good point, I'm happy to discuss it.

## A bit more context
First let me give more context to make sure we're on the same page: every OTA update targets a specific native version ("version" in the schema). So they usually reset every time it targets a new native version.
i.e. the versions of the javascript running in an app will be:
- version 1.0.0, bundle_version 1
- version 1.0.0, bundle_version 2
- version 1.0.0, bundle_version 3
- version 1.1.0, bundle_version 1

So we have to pass both informations to be able to identify crashes.

## Alternative solution
One alternative solution is to change the version to be something like `1.0.0:ota1` for version 1.0.0, bundle_version 1.
That way we don't have to add a new field, and everything that works for version still works here, so that looks like a good idea to me.

The current limitation we have is that it is not currently possible to set the version in the native SDKs.

Let me check with the rest of the mobile team if that is something that we can do.

louiszawadzki:
@BenoitZugmeyer First let me give more context to make sure we're on the same page: every OTA update targets a specific native version ("version" in the schema). So they usually reset every time it targets a new native version.
i.e. the versions of the javascript running in an app will be:

- version 1.0.0, bundle_version 1
- version 1.0.0, bundle_version 2
- version 1.0.0, bundle_version 3
- version 1.1.0, bundle_version 1

So we have to pass both informations to be able to identify crashes.
The SCI part is already done, we can send sourcemaps with a bundle_version and the unminification process will look for this to choose the correct sourcemap.

Good idea to colocate them, that makes sense to me.

schemas/rum/error-schema.json
ncreated: IMO it's better to leverage existing `version` and simply embed OTA version in it, like you propose in alternative solution @louiszawadzki. I'm only not sure if colon is safe, but dash definitely is (it's used in semver):
```
version = "<native app version>-ota<js bundle version>"
```
This way, every `version` will be uniquely determining the version of RN codebase (both the native and OTA code).

What I dislike about introducing `bundle_version` is that it leads to ambiguity. Both `version:1.0.0` and `bundle_version:3` queries will return results coming from more than one version of RN software. Given that `version` is very important to Datadog's [Unified Service Tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging) I'm not sure if we want to allow it.

mariusc83: I am actually on the same page with @ncreated on this one just for more visibility. You could also work with the `version` and `bundle_version` but having both values in one single property looks cleaner to me. Also you will only have one single facet this way which will reflect both versions for an app.

louiszawadzki: That makes sense to me.
I'm creating tasks to enable versions in SDKs.

We can still later add `bundle_version` if that is relevant later.

louiszawadzki:
Let's close this for now, we'll keep only one `version` attribute and add the possibility to set it to a custom value.

PR #78, opened 2022-07-28
Adjust the TextStyle mobile schema

mariusc83:
We are changing the way we provide the font family for the `TextWireframe` as this was not following the best practices and was adding an unjustified complexity. Furthermore the `type` property doesn't really make sense in mobile.

### DIFF
--- a/schemas/session-replay/mobile/text-style-schema.json
+++ b/schemas/session-replay/mobile/text-style-schema.json
@@ -6,22 +6,20 @@
   "description": "Schema of all properties of a TextStyle.",
   "allOf": [
     {
-      "required": ["color", "family", "size", "type"],
+      "required": [
+        "color",
+        "family",
+        "size"
+      ],
       "properties": {
         "family": {
           "type": "string",
-          "description": "The font family.",
+          "description": "The preferred font family collection, ordered by preference and formatted as a String list: e.g. Century Gothic, Verdana, sans-serif",
           "readOnly": true
         },
         "size": {
           "type": "integer",
           "description": "The font size in pixels.",
-          "readOnly": true
-        },
-        "type": {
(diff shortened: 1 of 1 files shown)


schemas/session-replay/mobile/text-style-schema.json
```diff
@@ -6,24 +6,23 @@
   "description": "Schema of all properties of a TextStyle.",
   "allOf": [
     {
-      "required": ["color", "family", "size", "type"],
+      "required": [
+        "color",
+        "family",
+        "size",
+        "type"
```
ncreated: There's no `type` now
```suggestion
```

ncreated (changes requested):
We need a fix in `reqired` attribtues;

schemas/session-replay/mobile/text-style-schema.json
```diff
@@ -6,24 +6,23 @@
...
+      ],
       "properties": {
         "family": {
           "type": "string",
-          "description": "The font family.",
+          "description": "The prefered font family collection ordered by preference and formatted as a String list: e.g. Century Gothic, Verdana, sans-serif",
```
acorretti: typo: `preferred`
suggestion: `... collection, ordered by preference...`

schemas/session-replay/mobile/text-style-schema.json
mquentin: Works for me, for now I leverage the size, family and color so there is no imapct on my side

mquentin (approved):

ncreated (changes requested):
It's still there 🙂 - required `type`, which makes the schema invalid.

ncreated (approved):
👌

PR #82, opened 2022-08-01
[REPLAY-946] Package rum-events-format repository and export typescript definitions

acorretti:
## What does this PR do?

* Declares a package.json
* Includes generated JS + TS definitions
* Configures prettier
* Prettifies all files
* Adds a CI job to ensure files are formatted
* Adds a CI job to ensure `yarn build` was ran with latest changes

## QA

* Tested CI jobs manually
* Ensured imports working as expected by referencing the package as a yarn git repository specifying a hash commit


package.json
```diff
@@ -0,0 +1,57 @@
...
+    "typescript": ">=4.5"
+  },
+  "engines": {
+    "node": ">=12.7.0"
+  },
+  "volta": {
+    "node": "16.3.0",
+    "yarn": "1.22.18"
+  }
+}
```
ThibautGeriz: **suggestion(not-blocking):** add line at the end of file (it's not covered by prettier?)

ThibautGeriz (commented):
Nicely done, it's not possible to generate some typescript file instead of .js and .d.ts ?

package.json
acorretti: Nice catch! I forgot to update the `yarn format` check to look at all files 😅 🙏

acorretti:
> Nicely done, it's not possible to generate some typescript file instead of .js and .d.ts ?

Even if not all entrypoints define JS constants –yet replay ones do– ESLint complains if an import doesn't have a corresponding JS entry file. This was pointed out by @Ayc0.

scripts/generate.js
```diff
@@ -0,0 +1,53 @@
...
+const SCHEMAS_PATH = 'schemas'
+const PRETTIER_CONFIG = '.prettierrc.yml'
+
+const DEFINITIONS = [
+  { source: 'rum-events-schema.json', name: 'rum' },
+  { source: 'telemetry-events-schema.json', name: 'telemetry' },
+  {
+    source: 'session-replay-browser-schema.json',
+    name: 'browserSessionReplay',
+    options: { additionalProperties: false },
```
BenoitZugmeyer, line 16: **💭 thought:** ‏ ‏we should have put `additionalProperties` in the `schema` files, so it is enforced for everyone

BenoitZugmeyer (approved):

README.md
```diff
@@ -22,17 +24,37 @@ https://jsonschema.net/
...
+
+Prerequisite:
+
+    yarn build:deps
+
+Generate:
+
+    yarn generate
+
+### Prettify TS & JSON files
```
bcaudan, line 66: **👏 praise:** ‏Nice 🎉

README.md
```diff
@@ -11,6 +11,8 @@ The format must stay backward compatible in order to be able to validate old eve
 
 Breaking changes to the format must be reflected by a major version update in `_dd.format_version` property.
 
+After editing any schema, you must re-generate the JS + TS definitions ([see below](#build-js-sources--ts-definitions)).
```
bcaudan, line 14: **💭 thought:** Did you consider generating those files only when installing this repository as a dependency?

scripts/generate.js
acorretti: I thought you guys were happy with them for both RUM & Telemetry events?
If it's not the case and you can ensure it all works fine, we could set this to `false` for all generated definitions. It's specific to the `json-schema-to-typescript` dependency, not to JSON-schema.

README.md
acorretti: The idea was for this repo to be consumed both as a dependency and as a submodule (browser-sdk) for the time being.
Also, in order to `yarn generate` you need first to `yarn build:deps` to get the _json-schema-to-typescript_ built, and that would happen for everyone developing on a consumer repo, at least once and every time we update the schemas.

ThibautGeriz (approved):

README.md
bcaudan: indeed

bcaudan (approved):

mquentin (approved):
LGTM ! really nice addition

jagracey (approved):

scripts/generate.js
BenoitZugmeyer: > I thought you guys were happy with them for both RUM & Telemetry events?

We are happy with additional properties on RUM (and Telemetry? I guess we didn't think much about it in this case) 🙂 

> If it's not the case and you can ensure it all works fine, we could set this to `false` for all generated definitions. It's specific to the `json-schema-to-typescript` dependency, not to JSON-schema.

In fact, it is [a JSON schema "keyword"](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties). We could add it to JSON-schema files so other teams could benefit from it if their JSON-schema library support it. One disadvantage is that it might be cumbersome to use it on every `object`...

acorretti: > In fact, it is [a JSON schema "keyword"](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties). We could add it to JSON-schema files so other teams could benefit from it if their JSON-schema library support it. One disadvantage is that it might be cumbersome to use it on every object...

Of course, that's the key we're using in [Session Replay's Attributes](https://github.com/DataDog/rum-events-format/blob/9e0f8b2ab26ce0e33b058d7d5d3663ab043847f5/schemas/session-replay/browser/attributes-schema.json?q=additionalProperties#L7) to declare the types for the `additionalProperties`.
Well because it'd be cumbersome to have to add it as `false` on every object, that's why I assumed you were meaning to define the _default_ value for it (what I did for the SessionReplay definitions generation).

Up to you. Could be a follow-up change as well.

PR #83, opened 2022-08-03
✨ Add `rule_psr` to RUM resource

liywjl:
As we are now passing `rule_psr` via the browser SDK to backend, it makes sense to add to here

`psr` = `Percentage Sample Rate`


schemas/rum/resource-schema.json
```diff
@@ -252,6 +252,12 @@
               "description": "trace identifier in decimal format",
               "pattern": "^[0-9]+$",
               "readOnly": true
+            },
+            "rule_psr": {
+              "type": "number",
+              "description": "sample rate in decimal format",
+              "pattern": "^[0-9]+$",
```
bcaudan: **❓ question:** ‏I guess we are not limited to integers, no?

schemas/rum/resource-schema.json
```diff
@@ -252,6 +252,12 @@
               "description": "trace identifier in decimal format",
               "pattern": "^[0-9]+$",
               "readOnly": true
+            },
+            "rule_psr": {
+              "type": "number",
+              "description": "sample rate in integer format",
+              "pattern": "^[1-9]$|^[1-9][0-9]$|^(100)$",
```
BenoitZugmeyer: **💬 suggestion:** ‏remove the `pattern` property here

I don't think the `pattern` keyword applies on `number` ([in the docs](https://json-schema.org/understanding-json-schema/reference/string.html#pattern) I only see it mentioned on `string` or object keys). Even if it does, I don't think checking the number format is useful here. Using a [range](https://json-schema.org/understanding-json-schema/reference/numeric.html?highlight=number#range) would be more appropriate.

schemas/rum/resource-schema.json
bcaudan: Looking at your change, I was probably not clear enough, I am thinking that we could provide values like `12,34` for example, so we should not limit this field to integer value.

liywjl: It should be clearer using the [range](https://json-schema.org/understanding-json-schema/reference/numeric.html?highlight=number#range) keys mentioned by @BenoitZugmeyer

ncreated (commented):
What does "PSR Rule" stand for? 🤔💭

xgouchet (approved):

liywjl:
> What does "PSR Rule" stand for? 🤔💭

I added the definition to the description

schemas/rum/resource-schema.json
```diff
@@ -242,6 +242,13 @@
               "description": "trace identifier in decimal format",
               "pattern": "^[0-9]+$",
               "readOnly": true
+            },
+            "rule_psr": {
+              "type": "number",
+              "description": "sample rate in integer format",
```
xgouchet: Maybe we need to be more precise as to which sampling rate this refer to. We have (at least) three: RUM session, RUM premium session and Tracing

schemas/rum/resource-schema.json
```diff
@@ -242,6 +242,13 @@
               "description": "trace identifier in decimal format",
               "pattern": "^[0-9]+$",
               "readOnly": true
+            },
+            "rule_psr": {
+              "type": "number",
+              "description": "tracing sample rate in integer format",
+              "minimum": 0,
+              "exclusiveMaximum": 100,
```
bcaudan: **💬 suggestion:** ‏Should it not be 1 then?

schemas/rum/resource-schema.json
```diff
@@ -242,6 +242,13 @@
               "description": "trace identifier in decimal format",
               "pattern": "^[0-9]+$",
               "readOnly": true
+            },
+            "rule_psr": {
+              "type": "number",
+              "description": "tracing sample rate in decimal format",
+              "minimum": 0,
+              "exclusiveMaximum": 1,
```
BenoitZugmeyer: **❓ question:** ‏I think this should be inclusive?


**💬 suggestion:** ‏
```suggestion
              "maximum": 1,
```

bcaudan (approved):

BenoitZugmeyer (approved):

PR #84, opened 2022-08-04
[REPLAY] MobileMutationData.source is mandatory

acorretti:
Same as for Browser, `MobileMutationData.source` must be mandatory to allow telling what has triggered an IncrementalRecord.

This will be needed in the future and enables interoperability in the player.

Additionally, `synchronize` is added as a `pull_request` trigger for CI actions.

### DIFF
--- a/dist/generated/rum.d.ts
+++ b/dist/generated/rum.d.ts
@@ -46,7 +46,7 @@ export declare type RumActionEvent = CommonProperties & {
             /**
              * Action frustration types
              */
-            readonly type: ('rage_click' | 'dead_click' | 'error_click')[];
+            readonly type: ('rage_click' | 'dead_click' | 'error_click' | 'rage_tap' | 'error_tap')[];
             [k: string]: unknown;
         };
         /**
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: ci
 on:
   push:
   pull_request:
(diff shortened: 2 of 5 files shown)


xgouchet (approved):

dist/generated/rum.d.ts
```diff
@@ -46,7 +46,7 @@ export declare type RumActionEvent = CommonProperties & {
             /**
              * Action frustration types
              */
-            readonly type: ('rage_click' | 'dead_click' | 'error_click')[];
+            readonly type: ('rage_click' | 'dead_click' | 'error_click' | 'rage_tap' | 'error_tap')[];
```
N-Boutaib, line 49: **Question:** Shouldn't we do the same for frustration records, since it's not shared ?

N-Boutaib (commented):
LGTM

dist/generated/rum.d.ts
acorretti: I only re-generated the definitions out of changes that were done previously in the schemas.

Maybe, in the future, but no need to do it now. It'll take some time until we see mobile recorders sending frustration records in replay payloads.

What you should ensure though, is for the player Frustration-handling logic to gracefully handle unknown frustration types.

BenoitZugmeyer (approved):

mariusc83 (approved):

N-Boutaib (approved):

mquentin (approved):
LGTM

PR #85, opened 2022-08-05
Build json-schema-to-typescript on demand

BenoitZugmeyer:
## Motivation

One less thing to manually install when using the tooling. Removes a bit of unneeded shell script.


liywjl (approved):

acorretti (approved):
🎉 nice!

PR #88, opened 2022-08-12
Rewrite schema validation script in JS

BenoitZugmeyer:
## Motivation

* This repo tend to have its tooling written in JS: let's unify this. This will simplify contributions.

* The previous script wasn't iterating [recursively over sample files](https://github.com/DataDog/rum-events-format/blob/fadc87171e6a477d8104daaaa908aa8e6f5e48b5/validate.sh#L21), which forced us to have [weird `json` directories](https://github.com/DataDog/rum-events-format/tree/fadc87171e6a477d8104daaaa908aa8e6f5e48b5/samples/session-replay-schema/mobile-segments/json). This could have been fixed, but [we are more comfortable in JS than in Bash](https://github.com/DataDog/rum-events-format/pull/65#discussion_r904820386)

Some schema changes were needed:

* Some schema ids were duplicated or incorrect. As Ajv rely on those id instead of filesystem path to identify schemas, they need to be unique and reflect the schema path.
* samples files are moved inside directories named after the schema they validate against
* some regex used a syntax unsupported by JS. Let's use syntaxes supported universally.


scripts/validate.js
```diff
@@ -0,0 +1,66 @@
...
+  if (valid) {
+    console.log(`✅ ${samplePath}`)
+  } else {
+    console.log(`❌ ${samplePath} is not valid against ${schemaId}:`)
+    console.log(`   - ${ajv.errorsText(undefined, { separator: '\n   - ' })}`)
+    process.exitCode = 1
+  }
+}
+
+function computeSchemaIdFromSchemaPath(schemaPath) {
```
liywjl, line 41: **❓ question:** Does it make sense to keep file extension `.json` for IDs?

liywjl (approved):

xgouchet (approved):
LGTM

bcaudan (approved):
LGTM

scripts/validate.js
BenoitZugmeyer: So I tried and technically it is possible to remove the extension from IDs, but because AJV resolves schemas `$ref` IDs, we need to change `$ref` as well. Other json-schema libraries might resolve `$ref` using file paths, so `$ref` should still represent a path name.

To keep compatibility with other libraries, let's keep using a real file path.

ThibautGeriz (approved):

PR #90, opened 2022-08-23
CommonJS + ESM package consumption readiness

acorretti:
## What does this PR do?

Generates both CommonJS and ESM targets for dual consumption.
Leverages _yarn workspace_ for sub-package definitions in order to emulate named exports for older tools.


xgouchet (approved):
LGTM

mbeaudru (approved):
Long comment sorry! First time in this repo, I'm discovering things all at once ahah

--------

Definition types are generated twice, both in ESM & CJS directories

From what I've seen, we also use the `generated` directory only to produce more definition types.

➡️  We could separate the definition types from the rest of the files, to have a `lib` directory like this:

```
lib/
  cjs/                    --> only contains cjs js files, without generated directory types and .d.ts
  esm/                 --> only contains esm js files, without generated directory types and .d.ts
  types/               --> only contains types
```

Having the .d.ts only once could slightly help vscode, single source of truth. But it's not blocker, if it's too hard for now, I won't RC and we can move on 👍 .

-----------

On a side note, we use the "fake trick" at multiple places right now with:

- "fake" packages to feed the `tsc`
- "fake" files to generate .d.ts
- "fake" file exports [export * from './generated/sessionReplay'](https://github.com/DataDog/rum-events-format/blob/93d1e2fc7d20f118bedeba0573ff730f1789de56/ts/session-replay.ts#L4)

I don't know how to feel about this - especially since I don't have a solution for point 2 and 3 on my mind.

Maybe at some point we should make those tricks explicit at least with some docs or comments :)

scripts/generate.mjs
```diff
@@ -72,7 +76,7 @@ async function getJsonSchemaToTypescript() {
       `
         set -eu
         cd ./node_modules/json-schema-to-typescript
-        rm -rf dist
+        rm -rf lib
```
BenoitZugmeyer: **🔨 warning:** json-schema-typescript did not change

```suggestion
        rm -rf dist
```

lib/generated/browserSessionReplay.ts
```diff
@@ -0,0 +1,636 @@
+/* eslint-disable */
```
BenoitZugmeyer: **🔨 warning:** remove `lib/generated/*` and adjust the `generated` folder path in gitignore

BenoitZugmeyer (approved):

PR #91, opened 2022-09-02
Add view.first_byte

bcaudan:
We already collect this data for resource events, cf [resource schema](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/resource-schema.json#L149).
Here the time to first byte of the resource corresponding to the document loading is attached to the view as a new timing metric.

More info on [ttfb](https://web.dev/ttfb/#measure-ttfb-in-javascript).

### DIFF
--- a/schemas/rum/view-schema.json
+++ b/schemas/rum/view-schema.json
@@ -103,6 +103,12 @@
               "minimum": 0,
               "readOnly": true
             },
+            "time_to_first_byte": {
+              "type": "integer",
+              "description": "Duration in ns to the response start of the document request",
+              "minimum": 0,
+              "readOnly": true
+            },
             "custom_timings": {
               "type": "object",
               "description": "User custom timings of the view. As timing name is used as facet path, it must contain only letters, digits, or the characters - _ . @ $",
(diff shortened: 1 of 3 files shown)


louisj-dd (approved):

xgouchet (approved):

schemas/rum/view-schema.json
```diff
@@ -103,6 +103,12 @@
               "minimum": 0,
               "readOnly": true
             },
+            "time_to_first_byte": {
```
BenoitZugmeyer: **💬 suggestion:** ‏
```suggestion
            "first_byte": {
```
* matches the resource naming
* matches other view namings where we don't use `time_to_...` (ex: we could've used `time_to_load_event` but we didn't)

bcaudan: ok sounds good to me it also match the original backend naming.
I'll do the change, cc @JoelPfaffDD

JoelPfaffDD: Please disregard the original backend naming choice I have taken, I needed a name to start the dev, and it is easy to change there.
Just consider the proposal above on its own merit, since this one cannot be changed easily anymore.

bcaudan: sure, @BenoitZugmeyer' arguments were enough for me 🙂

louisj-dd (approved):

liywjl (approved):

BenoitZugmeyer (approved):

PR #92, opened 2022-09-06
Changes to track new vitals using a common data format.

fuzzybinary:
The Mobile SDKs, especially the Cross Platform SDKs, want to add additional "vital" statistics. We're hoping to use a common format across these vitals to make adding more of them later easier, and could make creating a UI around vitals easier as well.

For more information, here's the RFC for supporting additional mobile vitals that uses a common format:
https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2535620778/RFC+-+Additional+Mobile+Vitals

But we're looking for feedback from the browser and backend teams as well on whether this format will work and be usable in all RUM technologies.


schemas/rum/_vital-info-schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "readOnly": true
+    },
+    "view_average": {
+      "type": "number",
+      "description": "The minimum value seen for this vital in the view.",
+      "readOnly": true
+    },
+    "vital_max": {
+      "type": "number",
+      "description": "The maximum value we could see for this vital.",
```
xgouchet: Is this `The maximum value we were able to see for this vital on this device since the first install of this app` or `The maximum value that could ever be seen for this vital on any device and any application`?

schemas/rum/_vital-info-schema.json
```diff
@@ -0,0 +1,30 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/_vital-info-schema.json",
+  "title": "RumVitalInfo",
+  "type": "object",
+  "description": "Schema of properties for a vital",
+  "required": ["view_min", "view_max", "view_average"],
+  "properties": {
+    "view_min": {
+      "type": "number",
+      "description": "The minimum value seen for this vital in the view.",
```
xgouchet: Nitpick: I'd replace `in the view` by `during this view's lifetime`
Same applies for `view_max` and `view_average`

xgouchet (approved):
LGTM

schemas/rum/view-schema.json
```diff
@@ -281,6 +281,24 @@
               "type": "number",
               "description": "Minimum refresh rate during the view’s lifetime (in frames per second)",
               "readOnly": true
+            },
+            "vitals": {
```
BenoitZugmeyer: AFAIK, `vitals` has been introduced by Google as a well defined metrics selection as well as threshold levels to deterministically evaluate the user experience of an application. Vitals should be reported as a single number easy to evaluate.

What you are sending are simple metrics. Calling them `vitals`may be confusing as it does not really match the same concept.


**💬 suggestion:** ‏I would suggest to treat them as other View metrics. As such, I would not nest them inside their own namepsace but add theme directly as `@view.*` attributes. You could still report multiple values (min/max/average). Example:

```json
{
  "view": {
    ...
    "flutter_build_time": {
      "min": 123,
      "max": 123,
      "avg": 123,
    }
}
```

schemas/rum/_vital-info-schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "type": "number",
+      "description": "The maximum value seen for this vital in the view.",
+      "readOnly": true
+    },
+    "view_average": {
+      "type": "number",
+      "description": "The minimum value seen for this vital in the view.",
+      "readOnly": true
+    },
+    "vital_max": {
```
BenoitZugmeyer: AFAIK, `vital_max` is not related to the view? It seems weird to include it in the View event.


**💬 suggestion:** ‏remove this value. Users can use the datadog UI to see the max value for all views from a session.

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 # Generate defintions out of schemas by running `yarn generate`
 lib/generated
+*/.DS_Store
```
BenoitZugmeyer, line 4: **🥜 nitpick:** ‏you might want to ignore .DS_Store files globally: https://dev.to/lpheller/quickly-ignore-dsstore-files-globally-ecd

schemas/rum/_vital-info-schema.json
fuzzybinary: This would be the maximum value possible for this session. This would be mostly for something like FPS, which can vary from device to device (and on iOS, sometimes can vary even within a session I believe....)

schemas/rum/_vital-info-schema.json
fuzzybinary: This was meant as "the maximum possible value" not the maximum we've seen. The use case generally is FPS on a mobile device, which can have a different max depending on the device. An app running at 59 FPS when the max is 60 is a very different metric from an app running at 59 FPS when the max is 120+, if that makes sense.

schemas/rum/view-schema.json
fuzzybinary: That might be a bigger conversation with @priyanshi-gupta -- we've been referring to these extra metrics as "Additional Mobile Vitals" but if we want to change the name to not confuse with Google's naming, I'm okay with that. Although "performance metrics" also has some baggage...

The hope with putting them into their own namespace was to (potentially) allow the front end to pick up these metrics and display them as "vitals" without having to know too much about them. Basically, anything in that namespace is assumed to be of the same type and displayed in the same way, queried the same way, etc. 

I'm okay moving them out of a namespace if it's not useful... just thought it might be.

schemas/rum/_vital-info-schema.json
BenoitZugmeyer: I see! Since it is related to the device, I would suggest to put it in the device field, ex: `@device.max_fps`

fuzzybinary: So, one idea for putting all this info together was to make creating UI for these metrics easier. The idea being that if we can come up with a generalized way of showing a "vital" (for lack of a better term) then all you need to know is the name of the property in the "vitals" object.

fuzzybinary:
Updated documentation based on review.

@BenoitZugmeyer Let me know if the "max" approach makes sense. I know we're weirdly sending a stat more for ease of front end implementation than for querying but I think it makes sense in this case?

schemas/rum/_perf-info-schema.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/_perf-info-schema.json",
+  "title": "RumPerfInfo",
+  "type": "object",
+  "description": "Schema of properties for a technical performance metic",
+  "required": ["view_min", "view_max", "view_average"],
+  "properties": {
+    "view_min": {
```
BenoitZugmeyer: **🥜 nitpick:** ‏maybe remove the `view_` prefix, as it is already in a `view` namespace I think it is already explicit enough.

BenoitZugmeyer (approved):

BenoitZugmeyer:
> Updated documentation based on review.
> 
> @BenoitZugmeyer Let me know if the "max" approach makes sense. I know we're weirdly sending a stat more for ease of front end implementation than for querying but I think it makes sense in this case?

Seems reasonable!

schemas/rum/_perf-info-schema.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/_perf-info-schema.json",
+  "title": "RumPerfInfo",
+  "type": "object",
+  "description": "Schema of properties for a technical performance metic",
```
bcaudan:
```suggestion
  "description": "Schema of properties for a technical performance metric",
```

schemas/rum/_perf-info-schema.json
```diff
@@ -0,0 +1,30 @@
...
+  "required": ["view_min", "view_max", "view_average"],
+  "properties": {
+    "view_min": {
+      "type": "number",
+      "description": "The minimum value seen for this metic during the view's lifetime.",
```
bcaudan:
```suggestion
      "description": "The minimum value seen for this metric during the view's lifetime.",
```

schemas/rum/_perf-info-schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "readOnly": true
+    },
+    "view_max": {
+      "type": "number",
+      "description": "The maximum value seen for this metic during the view's lifetime.",
```
bcaudan:
```suggestion
      "description": "The maximum value seen for this metric during the view's lifetime.",
```

schemas/rum/_perf-info-schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "readOnly": true
+    },
+    "view_average": {
+      "type": "number",
+      "description": "The minimum value seen for this metic during the view's lifetime.",
```
bcaudan:
```suggestion
      "description": "The average value for this metric during the view's lifetime.",
```

schemas/rum/_perf-info-schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "readOnly": true
+    },
+    "metric_max": {
+      "type": "number",
+      "description": "The maximum value we could see for this metic during this session.",
```
bcaudan:
```suggestion
      "description": "The maximum value we could see for this metric during this session.",
```

schemas/rum/_perf-info-schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "type": "number",
+      "description": "The maximum value seen for this metic during the view's lifetime.",
+      "readOnly": true
+    },
+    "view_average": {
+      "type": "number",
+      "description": "The minimum value seen for this metic during the view's lifetime.",
+      "readOnly": true
+    },
+    "metric_max": {
```
bcaudan: **💬 suggestion:** given the description, it could be more explicit and consistent with other field naming‏
```suggestion
    "session_max": {
```

schemas/rum/_perf-info-schema.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/_perf-info-schema.json",
+  "title": "RumPerfInfo",
```
bcaudan: **💬 suggestion:** ‏given the comments on this file, using metric directly could be more explicit
```suggestion
  "$id": "rum/_perf-metric-schema.json",
  "title": "RumPerfMetric",
```

schemas/rum/_perf-info-schema.json
fuzzybinary: Good suggestion! I'm still unhappy with this name though. It still sounds like this is the max we saw in the session, rather than the maximum possible value....

fuzzybinary:
@bcaudan Updated with most of your suggestions, can you take one more look through?

I kept `metric_max` but updated the documentation so it's a bit more clear what the purpose is.

schemas/rum/_perf-metric-schema.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/_perf-metric-schema.json",
+  "title": "RumPerfMetric",
+  "type": "object",
+  "description": "Schema of properties for a technical performance metic",
```
bcaudan:
```suggestion
  "description": "Schema of properties for a technical performance metric",
```

schemas/rum/_perf-metric-schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "readOnly": true
+    },
+    "average": {
+      "type": "number",
+      "description": "The average value for this metric during the view's lifetime.",
+      "readOnly": true
+    },
+    "metric_max": {
+      "type": "number",
+      "description": "The maximum value possible value we could see for this metric, if such a max is relevant and can vary from session to session.",
```
bcaudan:
```suggestion
      "description": "The maximum possible value we could see for this metric, if such a max is relevant and can vary from session to session.",
```

bcaudan (approved):
LGTM

PR #94, opened 2022-09-09
[RUMF-1352] Add causes field to error

liywjl:
Add [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) field to the error schema


schemas/rum/error-schema.json
```diff
@@ -47,6 +47,37 @@
...
+                  },
+                  "stack": {
+                    "type": "string",
+                    "description": "Stacktrace of the error",
+                    "readOnly": false
+                  },
+                  "origin": {
+                    "type": "object",
+                    "$ref": "error-schema.json"
+                  }
```
bcaudan: **❓ question:** ‏What is origin?

schemas/rum/error-schema.json
```diff
@@ -47,6 +47,37 @@
               "description": "Stacktrace of the error",
               "readOnly": false
             },
+            "causes": {
```
bcaudan, line 50: **💬 suggestion:** ‏It could be nice to add causes example in the [sample](https://github.com/DataDog/rum-events-format/blob/master/samples/rum-events/error.json)

schemas/rum/error-schema.json
```diff
@@ -47,6 +47,37 @@
...
+              "type": "array",
+              "description": "Causes of the error",
+              "readOnly": false,
+              "items": {
+                "type": "object",
+                "description": "Properties for one of the error causes",
+                "required": ["message"],
+                "properties": {
```
bcaudan, line 58: **❓ question:** ‏should not we have most of the other error properties?

schemas/rum/error-schema.json
liywjl: looks like `origin` was a mistake, instead it was meant to be `source`

xgouchet (approved):

schemas/rum/error-schema.json
```diff
@@ -47,6 +47,40 @@
               "description": "Stacktrace of the error",
               "readOnly": false
             },
+            "causes": {
+              "type": "array",
+              "description": "Causes of the error",
+              "maxItems": 10,
```
bcaudan: **💭 suggestion:** ‏the generated code seems to be quite verbose due to that. I'd be in favor of not putting this limit in the schema.

schemas/rum/error-schema.json
```diff
@@ -47,6 +47,40 @@
...
+              "readOnly": false,
+              "items": {
+                "type": "object",
+                "description": "Properties for one of the error causes",
+                "required": ["message"],
```
bcaudan: **💬 suggestion:** same as error, no?‏
```suggestion
                "required": ["message", "source"],
```

schemas/rum/error-schema.json
liywjl: 😅 wow good catch, yes I am up for removing it as a result

bcaudan (approved):
LGTM

BenoitZugmeyer (approved):

benoitmeriaux (approved):

PR #95, opened 2022-09-12
REPLAY-1033: Add BrowserSource and MobileSource consts

acorretti:
# Motivation

In order to have type guards for `SegmentMetadata['source']` properties, let's export some consts containing the possible values for both Browser & Mobile segments.

# Changes

Adds two exported objects: `BrowserSource` and `MobileSource`.

### DIFF
--- a/lib/src/session-replay-browser.ts
+++ b/lib/src/session-replay-browser.ts
@@ -2,6 +2,17 @@ import type * as SessionReplay from '../generated/browserSessionReplay'
 
 export * from '../generated/browserSessionReplay'
 
+/**
+ * For backward compatibility reasons, `undefined` should be accepted as a possible value for Browser segments, too.
+ **/
+export const BrowserSource: {
+  [key: string]: SessionReplay.BrowserSegmentMetadata['source']
+} = {
+  Browser: 'browser',
+}
+
+export type BrowserSource = typeof BrowserSource[keyof typeof BrowserSource]
+
 export const RecordType: {
   FullSnapshot: SessionReplay.BrowserFullSnapshotRecord['type']
   IncrementalSnapshot: SessionReplay.BrowserIncrementalSnapshotRecord['type']
--- a/lib/src/session-replay-mobile.ts
+++ b/lib/src/session-replay-mobile.ts
@@ -2,6 +2,17 @@ import type * as SessionReplay from '../generated/mobileSessionReplay'
 
 export * from '../generated/mobileSessionReplay'
 
+export const MobileSource: {
+  [key: string]: SessionReplay.MobileSegmentMetadata['source']
+} = {
+  Android: 'android',
+  Ios: 'ios',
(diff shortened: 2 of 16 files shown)


mquentin (approved):
LGTM thanks !

lib/src/session-replay-mobile.ts
```diff
@@ -2,6 +2,13 @@ import type * as SessionReplay from '../generated/mobileSessionReplay'
 
 export * from '../generated/mobileSessionReplay'
 
+export const MobileSource: SessionReplay.MobileSegmentMetadata['source'][] = [
+  'android',
+  'ios',
+  'flutter',
+  'react-native',
+]
+
```
BenoitZugmeyer: **💬 suggestion:** ‏Did you consider using the same pattern as other "enum"-like structures?

```suggestion
export const MobileSource = {
  Android: 'android',
  Ios: 'ios',
  Flutter: 'flutter',
  ReactNative: 'react-native'
} as const
export type MobileSource = typeof MobileSource[keyof typeof MobileSource]
```

mariusc83 (approved):

lib/src/session-replay-mobile.ts
BenoitZugmeyer: Seen on Slack: there is no point in using an object as those values will exclusively be used as an array.

lib/src/session-replay-browser.ts
```diff
@@ -2,6 +2,13 @@ import type * as SessionReplay from '../generated/browserSessionReplay'
 
 export * from '../generated/browserSessionReplay'
 
+export const BrowserSource: (undefined | SessionReplay.BrowserSegmentMetadata['source'])[] = [
+  /**
+   * For backward compatibility reasons, `undefined` should be accepted as a possible value for Browser segments, too.
+   **/
```
BenoitZugmeyer: **❓ question:** ‏did you intend to add `undefined` to this array then? Or is it something to check in web-ui?

BenoitZugmeyer (approved):

lib/src/session-replay-mobile.ts
acorretti: Eager to hear @mquentin 's take

lib/src/session-replay-browser.ts
acorretti: Well it's a consumer's responsibility, thus the comment. But the schema should be agnostic to compatibility issues.

xgouchet (approved):

lib/src/session-replay-mobile.ts
mquentin: So I agree that both approach would work but as a matter of consistency with the way we export `RecordType`, `IncrementalSource`... I would go for this change.

Notel: don't forget to do the same for `BrowserSource`

mquentin (approved):

PR #96, opened 2022-09-13
Add new pricing plans

bcaudan:
Update `_dd.session.plan` with new pricing codes

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -858,9 +858,9 @@ export interface CommonProperties {
          */
         session?: {
             /**
-             * Session plan: 1 is the 'lite' plan, 2 is the 'replay' plan
+             * Session plan: 1 is the 'lite' plan, 2 is the 'premium' plan, 3 is the 'pro' plan, 4 is the 'replay' plan
              */
-            plan: 1 | 2;
+            plan: 1 | 2 | 3 | 4;
             [k: string]: unknown;
         };
         /**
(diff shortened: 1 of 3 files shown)


BenoitZugmeyer (approved):

liywjl (approved):

xgouchet (approved):

lib/cjs/generated/rum.d.ts
```diff
@@ -858,9 +858,9 @@ export interface CommonProperties {
          */
         session?: {
             /**
-             * Session plan: 1 is the 'lite' plan, 2 is the 'replay' plan
+             * Session plan: 1 is the 'lite' plan, 2 is the 'premium' plan, 3 is the 'pro' plan, 4 is the 'replay' plan
              */
-            plan: 1 | 2;
+            plan: 1 | 2 | 3 | 4;
```
ThibautGeriz, line 863: **question:** replay was 2 and now it's 4, are we sure about this? It won't break existing behaviour?
**Disclaimer:** I have no idea what the new plans are

bcaudan: The only impact here is on the comment.
'replay' for 2 was outdated and we never fixed it to premium.
To be noted: 'pro' and 'replay' namings may still change, not a big deal for the format structure though.

acorretti (approved):

lib/cjs/generated/rum.d.ts
benoitmeriaux: 2 should be labeled as "premium" ? we never had this value in the backend 😅, it is "replay" everywhere, even in the "nicely formatted" field for customers

benoitmeriaux (approved):

lib/cjs/generated/rum.d.ts
bcaudan: it should have been but a bit late now 🙂

PR #104, opened 2022-09-28
[REPLAY] update Schemas to enable string ids for replay node map and mutations

mquentin:
## Context

There is no 1 to 1 mapping between wireframes and DOM nodes as 1 wireframe can be rendered with up to 3 nodes.

## Current implementation

currently the added nodes that are created while rendering a wireframe use unique number ids that need to be stored as metadata to later enable mutations.

## PR Improvements 

Instead of maintaining a map of these created number ids, using a string, this new ids can be prefixed and therefore `wireframe.id:1` can lead to `node.id:1`, `node.id:1-text-container` and `node.id:1-text-content`.

This allows us to apply the `TextMutation` of the `wireframe.id:1` as

```
TextMutation {
        id: '1-text-content',
        value: 'updatedText',
}
```

### DIFF
--- a/schemas/session-replay/browser/attribute-mutation-schema.json
+++ b/schemas/session-replay/browser/attribute-mutation-schema.json
@@ -7,7 +7,7 @@
   "required": ["id", "attributes"],
   "properties": {
     "id": {
-      "type": "integer",
+      "type": ["string", "number"],
       "description": "Id of the mutated node."
     },
     "attributes": {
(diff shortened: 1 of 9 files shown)


schemas/session-replay/browser/attribute-mutation-schema.json
```diff
@@ -7,7 +7,7 @@
   "required": ["id", "attributes"],
   "properties": {
     "id": {
-      "type": "integer",
+      "type": ["string", "number"],
```
mquentin, line 10: Should I rather go for number or integer ? (Mobile is mostly using number)

PR #105, opened 2022-10-05
Add configuration telemetry event

amortemousque:
Add the schema to collect customer configurations through telemetry events.
The goal is to have all configuration properties from all SDKs, so I added: 
- Common properties across all SDKs: `session_sample_rate`, `telemetry_sample_rate`, `trace_sample_rate`, `use_proxy`
- Browser SDK specialized properties mark
- Mobile SDK specialized properties mark


schemas/telemetry/configuration-schema.json
```diff
@@ -0,0 +1,125 @@
...
+                  "minimum": 0,
+                  "maximum": 100
+                },
+                "trace_sample_rate": {
+                  "type": "integer",
+                  "description": "The percentage of requests traced",
+                  "minimum": 0,
+                  "maximum": 100
+                },
+                "premium_sample_rate": {
```
bcaudan, line 46: **💬 suggestion:** ‏let's use session replay sample rate since premium is deprecated

schemas/telemetry/configuration-schema.json
```diff
@@ -0,0 +1,125 @@
...
+          "type": "object",
+          "description": "The telemetry configuration information",
+          "required": ["type", "configuration"],
+          "properties": {
+            "type": {
+              "type": "string",
+              "description": "Telemetry type",
+              "const": "configuration"
+            },
+            "configuration": {
```
bcaudan, line 24: **💬 suggestion:** ‏what about adding mobile specialized attributes as well?

amortemousque: My suggestion was to let each team add its specialised properties to be sure not to miss anything.
Wdyt?

bcaudan: Since some properties are already mentioned in the RFC, we could add them while we are here.
Maybe reach out to mobile team to see if they have a preference.

xgouchet (approved):
LGTM 👍

schemas/telemetry/configuration-schema.json
```diff
@@ -0,0 +1,186 @@
...
+                  "description": "Attribute to be used to name actions"
+                },
+                "use_allowed_tracing_origins": {
+                  "type": "boolean",
+                  "description": "Whether the allowed tracing origins list is used"
+                },
+                "default_privacy_level": {
+                  "type": "string",
+                  "description": "Session replay default privacy level",
+                  "enum": ["mask-user-input", "mask", "allow"]
```
bcaudan: **💭 thought:** I am wondering if listing the values here may not add more disturbance than value.
When adding new values, it would require us to also updates things here. 
Do you see a benefit for listing the allowed values here?
(same for other enum)
‏

amortemousque: Hum, it's true that maintaining these values can be painful. I'll remove them

bcaudan (approved):

RomanWuattier (approved):
LGTM. I'll create a task to add the default type `log` in the backend

PR #108, opened 2022-10-18
[RUM][REPLAY] Add shadow DOM metadata

ThibautGeriz:
Add support for shadow DOM in replay

Existing rrweb datastructure: https://github.com/rrweb-io/rrweb-snapshot/blob/775291042d8661af1a512832e2c3e44077466a4a/src/types.ts#L59-L60

Rrweb added two fields on the `Node` datastructure (HtmlElement, TextNode, Comment...): 
- `isShadowHost`: meaning when we build the node we have to call `node.attachShadow({ mode: 'open' })`
- `isShadow`: is the node a direct parent of a shadow DOM meaning when we build the node we have to add it to its parent using `node.shadowRoot.appendChild(childNode)` instead of `node.appendChild(childNode)`

We do need `isShadowHost` but I don't think we need `isShadow` because we already know the parent is a host.

The only possible issue that I have found is: what happens if we create a shadow root and add two types of children: normal child & shadow DOM. In this case, we see both tags in the DOM but only the shadow DOM is shown (see screenshot). For replay, we won't be recording the non-shadow DOM children. The rendering replay should be the same (but not the DOM). If we're not okay with that, we'll have to add the `isShadow` prop.

Screenshot 2022-10-19 at 09 39 44

### DIFF
--- a/schemas/session-replay/browser/element-node-schema.json
+++ b/schemas/session-replay/browser/element-node-schema.json
@@ -26,10 +26,12 @@
       }
     },
     "isSVG": {
-      "const": true
+      "const": true,
+      "description": "Is this node a SVG instead of a HTML"
     },
     "isShadowHost": {
-      "const": true
+      "const": true,
+      "description": "Is this node a host of a shadow root"
     }
   }
 }
--- a/schemas/session-replay/browser/element-node-schema.json
+++ b/schemas/session-replay/browser/element-node-schema.json
@@ -27,6 +27,9 @@
     },
     "isSVG": {
       "const": true
+    },
+    "isShadowHost": {
+      "const": true
     }
   }
 }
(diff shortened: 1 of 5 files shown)


mbeaudru:
> what happens if we create a shadow root and add two types of children: normal child & shadow DOM

When can this situation happen? From my understanding it would happen when the SDK wrongly flags DOM element as "not belonging in a shadow root" while it should, but I might be wrong.

If I'm right though, I think this is ok like this :)

acorretti (approved):

louisj-dd (approved):

bcaudan (approved):
LGTM

schemas/session-replay/browser/element-node-schema.json
```diff
@@ -27,6 +27,9 @@
     },
     "isSVG": {
       "const": true
+    },
+    "isShadowHost": {
+      "const": true
```
xgouchet: Can you add a description on this new attribute please?

xgouchet (approved):

mquentin (approved):
LGTM thanks for the ADR

mbeaudru (approved):

PR #109, opened 2022-10-21
add nullable clip field

mquentin:
`clip` is optional (some wireframes don’t have any clip).

If you send an update that remove the clipping, either you need to send a `clip:{top:0, right:0, left:0, bottom:0,}` or a `clip: null` if you want to replace the wireframe clipping stored in the player memory.

The second option makes more sense to me.

EDIT: Mobile sdk would have trouble to send `clip: null` on their end. we will go for the other approach.

### DIFF
--- a/schemas/session-replay/mobile/wireframe-clip-schema.json
+++ b/schemas/session-replay/mobile/wireframe-clip-schema.json
@@ -2,7 +2,7 @@
   "$schema": "http://json-schema.org/draft-07/schema",
   "$id": "session-replay/mobile/wireframe-clip-schema.json",
   "title": "WireframeClip",
-  "type": "object",
+  "type": ["object", "null"],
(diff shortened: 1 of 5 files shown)


schemas/session-replay/mobile/wireframe-clip-schema.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "http://json-schema.org/draft-07/schema",
   "$id": "session-replay/mobile/wireframe-clip-schema.json",
   "title": "WireframeClip",
-  "type": "object",
+  "type": ["object", "null"],
```
mariusc83, line 5: This will create issues on our end and we will have to update the schema generators. I want to challenge this as we never used this approach before. Is this a correct way of declaring types in a JSON schema ?

mariusc83 (changes requested):

PR #111, opened 2022-10-25
Add resolveDuration to RumResourceEvent

liywjl:
We want to track the timings that are registered when the `fetch` resource are resolved

### DIFF
--- a/schemas/rum/resource-schema.json
+++ b/schemas/rum/resource-schema.json
@@ -60,6 +60,12 @@
               "minimum": 0,
               "readOnly": true
             },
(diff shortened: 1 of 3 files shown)


schemas/rum/resource-schema.json
```diff
@@ -60,6 +60,12 @@
               "minimum": 0,
               "readOnly": true
             },
+            "resolveDuration": {
```
bcaudan, line 63: **❓ question:** ‏is it something that we plan to keep?
if not, let's not expose it to customer and put it on `_dd`

PR #112, opened 2022-10-27
Add Mobile specific attributes for Configuration Telemetry event

xgouchet:


schemas/telemetry/configuration-schema.json
```diff
@@ -24,6 +24,7 @@
             "configuration": {
               "type": "object",
               "description": "Configuration properties",
+              "additionalProperties": true,
```
ncreated: Why `additionalProperties: true` 🤔💭? Do we want to allow arbitrary configuration fields?

schemas/telemetry/configuration-schema.json
```diff
@@ -178,6 +179,48 @@
...
+                "batch_size": {
+                  "type": "string",
+                  "description": "The size of batches sent by the SDK",
+                  "enum": ["small", "medium", "large"]
+                },
+                "batch_upload_frequency": {
+                  "type": "string",
+                  "description": "The upload frequency of batches",
+                  "enum": ["frequent", "average", "rare"]
                 }
```
ncreated, line 229: We're missing some fields compared to the list in _RFC - Mobile Configuration Telemetry_:

- `track_native_long_tasks`
- `track_cross_platform_long_tasks`
- `track_network_requests`
- `use_tracing`

lib/esm/generated/telemetry.d.ts
```diff
@@ -195,6 +195,46 @@ export declare type TelemetryConfigurationEvent = CommonTelemetryProperties & {
...
+             */
+            track_flutter_performance?: boolean;
+            /**
+             * The size of batches sent by the SDK
+             */
+            batch_size?: 'small' | 'medium' | 'large';
+            /**
+             * The upload frequency of batches
+             */
+            batch_upload_frequency?: 'frequent' | 'average' | 'rare';
```
louiszawadzki: I assume that if the user sets it to `never`, we leave this field empty, is it the case?
If so, do we also leave it empty if no value is explicitly specified by the user, or do we put the default value?

In both cases we can have 2 different configurations that translate into the same reported values.
In my opinion it could be worth specifying `'rare'` here, if for instance we want to change the default value that would bring useful information.

schemas/telemetry/configuration-schema.json
```diff
@@ -178,6 +179,48 @@
...
+                },
+                "track_errors": {
+                  "type": "boolean",
+                  "description": "Whether error monitoring & crash reporting is enabled for the source platform"
+                },
+                "track_native_views": {
+                  "type": "boolean",
+                  "description": "Whether native views are tracked (for cross platform SDKs)"
+                },
+                "track_native_errors": {
```
louiszawadzki, line 194: We already have a `track_native_crashes` key (`"Whether native crashes are tracked"`), I think this one is a duplicate?

schemas/telemetry/configuration-schema.json
louiszawadzki: What is the impact of reporting arbitrary fields?
Are they going to be treated just like the ones that are declared?

I find it convenient for temporary fields, when we want to monitor deprecated fields usage for instance, or when we want to deprecate certain versions of a dependency (like React Native or React).
The risk of adding those in the schema is that we forget to remove them and it becomes bloated after a few months, and it also increases a little the time to be able to report it.

ncreated: I just wonder where is the line between reporting something as a custom property vs tracking it in the schema. If it's not about "platform-specifics" (we have Flutter- or Android-specific fields), then what else? The example of deprecated configuration makes sense and I like it 👍, but I believe that using `additionalProperties` should be followed with some agreed convention, to avoid ambiguity.

schemas/telemetry/configuration-schema.json
```diff
@@ -178,6 +179,48 @@
...
+                },
+                "batch_size": {
+                  "type": "string",
+                  "description": "The size of batches sent by the SDK",
+                  "enum": ["small", "medium", "large"]
+                },
+                "batch_upload_frequency": {
+                  "type": "string",
+                  "description": "The upload frequency of batches",
+                  "enum": ["frequent", "average", "rare"]
```
plousada:
```suggestion
                  "type": "integer",
                  "description": "The upload frequency of batches (in milliseconds)"
```
We should change this to match mobile vitals frequency as well.

schemas/telemetry/configuration-schema.json
```diff
@@ -178,6 +179,48 @@
...
+                  "description": "The type of initialization the SDK used, in case multiple are supported"
+                },
+                "track_flutter_performance": {
+                  "type": "boolean",
+                  "description": "Whether Flutter build and raster time tracking is enabled"
+                },
+                "batch_size": {
+                  "type": "string",
+                  "description": "The size of batches sent by the SDK",
+                  "enum": ["small", "medium", "large"]
```
plousada:
```suggestion
                  "type": "integer",
                  "description": "The time window considered for creating batches (in milliseconds)"
```
In order to better match types of mobile vitals frequency and batch upload frequency.

plousada (changes requested):

lib/esm/generated/telemetry.d.ts
xgouchet: We don't have an option to set it to 'never', and we'll change everything to report durations in ms

schemas/telemetry/configuration-schema.json
xgouchet: The idea is to allow for very specific attributes that we might want to add to a configuration event, but not necessarily to the schema. But agreed it seems we want to list all the fields here, I'll remove it.

schemas/telemetry/configuration-schema.json
xgouchet: This is what happens when the RFC moves in the middle of implementation 😅

schemas/telemetry/configuration-schema.json
xgouchet: Yep the RFC changed after I sent the PR 😅

schemas/telemetry/configuration-schema.json
xgouchet: Yep the RFC changed after I sent the PR 😅

ncreated (commented):
Recent additions look good 👍 although it's still missing the `use_tracing` field from RFC. Shouldn't be there?

BenoitZugmeyer (approved):

xgouchet:
> Recent additions look good 👍 although it's still missing the `use_tracing` field from RFC. Shouldn't be there?

Well I asked a question on the RFC, not sure what's the difference between `use_tracing` and `track_network_requests`.

schemas/telemetry/configuration-schema.json
```diff
@@ -175,9 +175,61 @@
                   "type": "integer",
                   "description": "The period between each Mobile Vital sample (in milliseconds)"
                 },
-                "track_native_crashes": {
+                "track_actions": {
                   "type": "boolean",
-                  "description": "Whether native crashes are tracked"
+                  "description": "Whether action tracking is performed automatically"
+                },
```
bcaudan: Could it be the same as `trackInteractions`?

schemas/telemetry/configuration-schema.json
```diff
@@ -175,9 +175,61 @@
...
+                  "description": "Whether error monitoring & crash reporting is enabled for the source platform"
+                },
+                "track_network_requests": {
+                  "type": "boolean",
+                  "description": "Whether tracing features are enabled"
+                },
+                "use_tracing": {
+                  "type": "boolean",
+                  "description": "Whether the cross-platform SDK was initialized on top of a pre-existing native SDK instance"
+                },
```
bcaudan: The description does not refer to tracing, is it normal?

schemas/telemetry/configuration-schema.json
```diff
@@ -175,9 +175,61 @@
...
+                  "description": "Whether long task tracking is performed automatically for cross platform SDKs"
+                },
+                "use_attach_to_existing": {
+                  "type": "boolean",
+                  "description": "Whether the cross-platform SDK was initialized on top of a pre-existing native SDK instance"
+                },
+                "use_first_party_hosts": {
+                  "type": "boolean",
+                  "description": "Whether the client has provided a list of first party hosts"
+                },
```
bcaudan, line 213: Could it be the same as `use_allowed_tracing_origins`?

schemas/telemetry/configuration-schema.json
xgouchet: Nice catch, I ended up modifying the wrong one

schemas/telemetry/configuration-schema.json
xgouchet: No, first party hosts have other impacts than just tracing

ncreated (approved):
LG 👌💯!

RomanWuattier (approved):

lib/esm/generated/telemetry.d.ts
louiszawadzki: oh yes sorry, I mixed it with some other fields

louiszawadzki (approved):

plousada (approved):

schemas/telemetry/configuration-schema.json
xgouchet: done

bcaudan (approved):

PR #113, opened 2022-11-08
[RUMF-1402] Add feature_flags to view to error

amortemousque:
In the context of Feature Flag Analysis, we need to be able to collect them from RUM SDKs. 
This PR adds the `feature_flags` attribute to `view` and `error`

### DIFF
--- a/schemas/rum/error-schema.json
+++ b/schemas/rum/error-schema.json
@@ -189,7 +189,7 @@
         "feature_flags": {
           "type": "object",
           "description": "Feature flags properties",
-          "additionalProperties": { "type": ["boolean", "string"] },
+          "additionalProperties": true,
           "readOnly": true
         }
       }
--- a/schemas/rum/error-schema.json
+++ b/schemas/rum/error-schema.json
@@ -185,6 +185,12 @@
             }
           },
           "readOnly": true
+        },
+        "feature_flags": {
+          "type": "object",
+          "description": "Feature flags properties",
+          "additionalProperties": { "type": ["boolean", "string"] },
+          "readOnly": true
         }
       }
     }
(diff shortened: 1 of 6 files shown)


liywjl (approved):

schemas/rum/error-schema.json
```diff
@@ -185,6 +185,12 @@
             }
           },
           "readOnly": true
+        },
+        "feature_flags": {
+          "type": "object",
+          "description": "Feature flags properties",
+          "additionalProperties": { "type": ["boolean", "string"] },
```
bcaudan: **💭 thought:** ‏any hard reason to limit to boolean and string? from brief, we can see mentions of number, null, json

amortemousque: - JSON values : From the backend RFC, in the requirement section they discard json from the first iteration. 
- numbers: The backend RFC only mention string and boolean but I'll sync with them to clarify it
- null values: From the Nazim RFC, we only collect feature flag used so we won't have null values

bcaudan: > null values: From the Nazim RFC, we only collect feature flag used so we won't have null values

FMU, Nazim's RFC is about integrations, we could have customers using the API directly and provide us a wide variety of inputs.
By limiting types here, we would need to enforce it by casting or rejecting inputs. 
So I am wondering if there is a real need or if we could be a bit more permissive.

xgouchet (approved):
LGTM

schemas/rum/error-schema.json
amortemousque: @Miz85

Miz85: Hi, for context we decided to leave JSON values out of the equation for V0 in order to simplify the work and to avoid hitting size limitations that might exist on the backend side (and maybe on the SDK side?). The thought process was that even if you could technically stringify a JSON and store it in a string, regular usage would be using a JSON value to store JSON on the Feature Flag Provider side.

We do want to support them ultimately though and talking with @joaooliveira-datadog it seems that the backend will drop feature flags for which the format is not supported so if it's less work not to filter them for now and you don't think it's risks having performance impacts on the SDK then I don't see a reason not to do it. The backend will pick them up when the support for it will be implemented.

Does that answer your concerns?

PR #114, opened 2022-11-08
[RUMF-1432] Add `track_resources` and `track_long_task`

liywjl:
Add `track_resources` and `track_long_task` to track for configuration telemetry

### DIFF
--- a/lib/esm/generated/telemetry.d.ts
+++ b/lib/esm/generated/telemetry.d.ts
@@ -128,6 +128,14 @@ export declare type TelemetryConfigurationEvent = CommonTelemetryProperties & {
              */
             track_session_across_subdomains?: boolean;
             /**
+             * Whether resources are tracked
+             */
+            track_resources?: boolean;
+            /**
+             * Whether long tasks are tracked
+             */
+            track_long_task?: boolean;
+            /**
              * Whether a secure cross-site session cookie is used
              */
             use_cross_site_session_cookie?: boolean;
--- a/lib/cjs/generated/telemetry.d.ts
+++ b/lib/cjs/generated/telemetry.d.ts
@@ -128,6 +128,14 @@ export declare type TelemetryConfigurationEvent = CommonTelemetryProperties & {
              */
             track_session_across_subdomains?: boolean;
             /**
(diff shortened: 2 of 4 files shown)


lib/esm/generated/telemetry.d.ts
```diff
@@ -127,6 +127,14 @@ export declare type TelemetryConfigurationEvent = CommonTelemetryProperties & {
...
              */
             track_session_across_subdomains?: boolean;
+            /**
+             * Whether resources are tracked
+             */
+            track_resources?: boolean;
+            /**
+             * Whether long tasks are tracked
+             */
+            track_long_task?: boolean;
```
louiszawadzki, line 137: 1. on iOS, Android and React Native, we can also specify a threshold for long tasks. Do we want to also track this? - on RN we have a threshold for javascript and native long tasks

2. (maybe for another PR) on React Native we can track javascript long tasks, and also native long tasks. I realise that we usually follow a pattern where 
- `track_X` means "track javascript X"
- `track_native_X` means "track native X on cross-platform"
Now, for long tasks we have `track_cross_platform_long_tasks` which follows the opposite pattern 😢 
Maybe this is the right time to change `track_cross_platform_long_tasks` to `track_native_long_tasks`. 
If you think this can create more discussion, I can open a separate PR for this change

amortemousque, line 137: **💬 suggestion:** ‏You may want to add these new attributes to the configuration samples used for validation

xgouchet (approved):

lib/esm/generated/telemetry.d.ts
liywjl: 1 - We do not use thresholds nor specify sampling rate/thresholds per metric on the browser SDK. We currently only have sampling rate that applies to all metrics and then booleans to disable specific metrics from being collected. It is an interesting idea to explore however, so I have reached out to the team to discuss it

2 - Great point! I think it makes sense as a different PR as this would effect mainly mobile SDKs and the aim here was just to add the missing flags of browser SDK.

amortemousque (approved):

PR #115, opened 2022-11-11
Set properties that potentially "late init" to mutable

fuzzybinary:
The Cross Platform mobile SDKs have certain telemetry properties that cannot be known until after the call for initialization. Before we send the telemetry configuration, we need these SDKs to be able to modify certain properties before we send them to telemetry. 

I also removed a Flutter specific property 'attachToExisting' that cannot be known early enough in the app lifecycle to be sent with telemetry.

### DIFF
--- a/schemas/telemetry/configuration-schema.json
+++ b/schemas/telemetry/configuration-schema.json
@@ -65,11 +65,13 @@
                   "type": "integer",
                   "description": "The percentage of sessions with Browser RUM & Session Replay pricing tracked",
                   "minimum": 0,
-                  "maximum": 100
+                  "maximum": 100,
+                  "readOnly": false
                 },
                 "use_proxy": {
                   "type": "boolean",
-                  "description": "Whether a proxy configured is used"
+                  "description": "Whether a proxy configured is used",
+                  "readOnly": false
                 },
                 "use_before_send": {
                   "type": "boolean",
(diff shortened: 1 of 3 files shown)


schemas/telemetry/configuration-schema.json
```diff
@@ -65,11 +65,13 @@
                   "type": "integer",
                   "description": "The percentage of sessions with Browser RUM & Session Replay pricing tracked",
                   "minimum": 0,
-                  "maximum": 100
+                  "maximum": 100,
+                  "readOnly": false
```
plousada, line 69: Why change `session_replay_sample_rate`?

schemas/telemetry/configuration-schema.json
```diff
@@ -101,7 +103,8 @@
                 },
                 "default_privacy_level": {
                   "type": "string",
-                  "description": "Session replay default privacy level"
+                  "description": "Session replay default privacy level",
+                  "readOnly": false
```
plousada, line 107: Same question.

plousada (changes requested):

schemas/telemetry/configuration-schema.json
fuzzybinary: The thinking was that Flutter's implementation of Session Replay will likely be completely independent of native session replay. I won't be sending down the sampling rate or privacy level to the native configuration objects, so I'll need to supply the values from the Flutter plugin.

xgouchet (approved):

plousada (approved):

BenoitZugmeyer (approved):

PR #116, opened 2022-11-22
[RUM][session replay] add PointerInteraction schemas

mquentin:
### Context

After investigating the current `TouchData` limitations, a new Mobile/Browser agnostic interaction schema will be added: `PointerInteraction`.

### Changes

- Add `PointerInteraction` related schemas
- Add the new `PointerInteractionData` to the list of existing `MobileIncrementalData` and `BrowserIncrementalData`
- Add the TS constant in view to be able to consume it web-ui.

### Note

`PointerInteractionData` is added to the `BrowserIncrementalData` even tho it will not be supported yet to prevent any conflicts with other potential  `BrowserIncrementalData` additions. As a result `PointerInteractionData` will be `source:9`

### QA

[Browser schemat] Validate that `PointerInteractionData` `source:9` has no conflicts with existing `BrowserInteractionData` sources


schemas/session-replay/browser/incremental-data-schema.json
```diff
@@ -28,6 +28,9 @@
     },
     {
       "$ref": "../common/viewport-resize-data-schema.json"
+    },
+    {
+      "$ref": "../common/pointer-interaction-data-schema.json"
```
mquentin, line 33: `PointerInteractionData` is added for browser even tho it will not be supported yet to prevent any conflicts with other potential `BrowserIncrementalData` additions. As a result `PointerInteractionData` will be `source:9`

schemas/session-replay/common/pointer-event-type-schema.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/common/pointer-event-type-schema.json",
+  "title": "PointerEventType",
+  "description": "Schema of a PointerEventType",
+  "type": "integer",
+  "enum": [0, 1, 2],
```
mquentin: cf.

```
export declare const PointerEventType: {
    readonly PointerDown: 0;
    readonly PointerUp: 1;
    readonly PointerMove: 2;
};
```

schemas/session-replay/common/pointer-interaction-data-schema.json
```diff
@@ -0,0 +1,23 @@
...
+  "title": "PointerInteractionData",
+  "type": "object",
+  "description": "Schema of a PointerInteractionData.",
+  "allOf": [
+    {
+      "required": ["source"],
+      "properties": {
+        "source": {
+          "type": "integer",
+          "const": 9,
```
mquentin, line 13: Set to `9` to bump the highest source from `schemas/session-replay/browser/incremental-data-schema.json`

schemas/session-replay/common/pointer/pointer-event-type-schema.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/common/pointer/pointer-event-type-schema.json",
+  "title": "PointerEventType",
+  "description": "Schema of a PointerEventType",
+  "readOnly": true,
+  "oneOf": [
+    {
+      "$ref": "pointer-down-schema.json"
```
mariusc83: Instead of using dedicated types for these I would rather use enums and even ints with a description for each int property. This will save us some space and also will be more optimised when having to serialize the events in the mobile sdk. We are already experiencing some issues with this in Android SDK and would like whenever is possible to simplify the models a bit. @0xnm

schemas/session-replay/common/pointer/pointer-type-schema.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/common/pointer/pointer-type-schema.json",
+  "title": "PointerType",
+  "description": "Schema of an PointerType",
+  "readOnly": true,
+  "oneOf": [
```
mariusc83: Same for these ones,  I would rather have them as enums.

schemas/session-replay/common/pointer/pointer-event-type-schema.json
mquentin: Ok after a discussion with Benoît, it seems that the enum is a good approach:
I will use string values for the [pointer-event-type-schema.json](https://github.com/DataDog/rum-events-format/pull/116/files/a482e7ac419971d17d6889a98517b9b5d875ce64#diff-fe73ab245b110f6ec5345524eae5ed135e929051a24970a06ac7633cf12c1c6a) as it is more intuitive as integers.

Note that the downside of this approach is that we will have to introduce a constant `PointerEventType` as

```
export const PointerEventType: {
  PointerDown: string,
  PointerUp: string,
  PointerMove: string
} = {
  PointerDown: 'down',
  PointerUp: 'up',
  PointerMove: 'move'
} as const
```

That will not consume the generated type as the following example `BrowserFullSnapshot: typeof BrowserRecordType.FullSnapshot` in

```
export const RecordType: {
  BrowserFullSnapshot: typeof BrowserRecordType.FullSnapshot
  BrowserIncrementalSnapshot: typeof BrowserRecordType.IncrementalSnapshot
  ...
} = {
  BrowserFullSnapshot: 2,
  BrowserIncrementalSnapshot: 3,
  ...
} as const
```

lib/src/session-replay.ts
```diff
@@ -43,3 +43,27 @@ export const RecordType: {
 export type RecordType = typeof RecordType[keyof typeof RecordType]
 
 export type IncrementalSource = BrowserIncrementalSource | MobileIncrementalSource
+
+export const PointerEventType: {
```
mquentin, line 47: related to this comment:
https://github.com/DataDog/rum-events-format/pull/116#discussion_r1030253458

the enum approach prevent a `PointerDown: typeof PointerEventType.Down` approach.

As discussed with Benoît it is good enough like this as the string enums are explicit enough.

fuzzybinary (approved):

mariusc83 (approved):
Looks way better. Tks !!

acorretti (approved):

PR #118, opened 2022-11-24
[RUM] add new incrementalsource to the exported const

mquentin:
### Change

add missing `PointerInteractionData` source to the existing 
`BrowserIncrementalSource` and `MobileIncrementalSource`


BenoitZugmeyer (approved):
Looks fine. You could even remove the font comment as it was never implemented in the SDK

PR #120, opened 2022-12-02
Remove invalid property from `segment-metadata-schema`

ncreated:
### What & Why?

🐞 This PR fixes `segment-metadata-schema.json` for mobile - it was requiring `has_full_snapshot_record` property, but it wasn't defining its schema.

### How?

I'm removing `has_full_snapshot_record` from the list of required properties. If we really need it, it should be added back along with its schema definition.

### DIFF
--- a/schemas/session-replay/mobile/segment-metadata-schema.json
+++ b/schemas/session-replay/mobile/segment-metadata-schema.json
@@ -12,7 +12,7 @@
       "$ref": "../common/_common-segment-metadata-schema.json"
     },
     {
-      "required": ["source", "has_full_snapshot_record"],
+      "required": ["source"],
       "properties": {
         "source": {
           "type": "string",


ThibautGeriz (approved):
Indeed, the property is still defined in `schemas/session-replay/common/_common-segment-metadata-schema.json` and it's not required

acorretti:
> Indeed, the property is still defined in `schemas/session-replay/common/_common-segment-metadata-schema.json` and it's not required

True, but if we remove it entirely –not a bad idea– then @mquentin should remove its usage from `SessionReplayDebug.tsx` and the fixtures, in web-ui.

ThibautGeriz:
> > Indeed, the property is still defined in `schemas/session-replay/common/_common-segment-metadata-schema.json` and it's not required
> 
> True, but if we remove it entirely –not a bad idea– then @mquentin should remove its usage from `SessionReplayDebug.tsx` and the fixtures, in web-ui.

It's still gonna be here in mobile & browser, it's fine

PR #125, opened 2022-12-21
[REPLAY] Add support dynamic stylesheet

ThibautGeriz:
📺 Replay only
🕸️ Browser only

MDN links:
- https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets
- https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/adoptedStyleSheets
- https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/CSSStyleSheet

image

### DIFF
--- a/lib/cjs/generated/browserSessionReplay.d.ts
+++ b/lib/cjs/generated/browserSessionReplay.d.ts
@@ -370,7 +370,7 @@ export interface StyleSheet {
     /**
      * CSS rules applied (rule.cssText)
      */
-    rules: string[];
+    cssRules: string[];
     /**
      * MediaList of the stylesheet
      */
--- a/lib/cjs/generated/browserSessionReplay.d.ts
+++ b/lib/cjs/generated/browserSessionReplay.d.ts
@@ -357,7 +357,28 @@ export interface DocumentNode {
      * The type of this Node.
      */
     readonly type: 0;
+    /**
+     * Stylesheet added dynamically
+     */
+    readonly adoptedStyleSheets?: StyleSheet[];
     childNodes: SerializedNodeWithId[];
+}
+/**
+ * Browser-specific. Schema of a StyleSheet
+ */
+export interface StyleSheet {
+    /**
+     * CSS rules applied (rule.cssText)
+     */
+    rules: string[];
+    /**
+     * MediaList of the stylesheet
+     */
+    media?: string[];
+    /**
+     * Is the stylesheet disabled
+     */
+    disabled?: boolean;
 }
 /**
  * Schema of a Document FragmentNode.
@@ -367,6 +388,10 @@ export interface DocumentFragmentNode {
      * The type of this Node.
      */
     readonly type: 11;
+    /**
+     * Stylesheet added dynamically
+     */
+    readonly adoptedStyleSheets?: StyleSheet[];
     /**
      * Is this node a shadow root or not
      */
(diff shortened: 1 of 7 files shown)


schemas/session-replay/browser/stylesheet-schema.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/browser/stylesheet-schema.json",
+  "title": "StyleSheet",
+  "type": "object",
+  "description": "Browser-specific. Schema of a StyleSheet",
+  "required": ["rules"],
+  "properties": {
+    "rules": {
```
yannickadam: MDN mentions that `rules` is a legacy property, replaced by `cssRules`: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet#legacy_properties
Maybe we want to match this?

lib/cjs/generated/browserSessionReplay.d.ts
```diff
@@ -357,8 +357,29 @@ export interface DocumentNode {
...
     childNodes: SerializedNodeWithId[];
 }
+/**
+ * Browser-specific. Schema of a StyleSheet
+ */
+export interface StyleSheet {
+    /**
+     * CSS rules applied (rule.cssText)
+     */
+    rules: string[];
```
BenoitZugmeyer: **❓ question:** ‏‏for inline stylesheets, we serialize all rules as a big `_cssText` string attribute. Any reason to use an array here? I'd be fine with it, but bear in mind that some rules can have child rules, so we could still have multiple rules in a single "rule" entry.

BenoitZugmeyer (commented):
* Ahve you plans to introduce a IncrementalMutation triggered when the list changes later? (ex: `document.adoptedStyleSheets = [newSheet]`)
* Are you planning to support adopted StyleSheet mutations later ? (ex: `newSheet.addRule('...')`)

lib/cjs/generated/browserSessionReplay.d.ts
ThibautGeriz: Indeed, it make sense to reuse `getCssRulesString` in the SDK

ThibautGeriz:
>    Ahve you plans to introduce a IncrementalMutation triggered when the list changes later? (ex: document.adoptedStyleSheets = [newSheet])
    Are you planning to support adopted StyleSheet mutations later ? (ex: newSheet.addRule('...'))

Honestly I am not planing to, I am hopeful that it won't be necessary to support shadow dom in practice but I agree it should be supported at some point. 
That being said, I will quickly QA the second point and see if it's a quick win. If it is then I will support it otherwise, I will be a ticket in JIRA

lib/cjs/generated/browserSessionReplay.d.ts
ThibautGeriz: we chatted, we'll keep the array of string to avoid using `replaceSync` that has a not so good support in the browser or using a hack with `insertRule`

mariusc83 (approved):

mbeaudru (approved):

acorretti (approved):

JoelPfaffDD (approved):
👍

BenoitZugmeyer (approved):

yannickadam (approved):

PR #127, opened 2023-01-20
[RUM] introduce image wireframe schema

mquentin:
# Motivation

Introduce new `ImageWireframe` schema to `[rum-events-format](https://github.com/DataDog/rum-events-format)` 

# Changes

- Add  `image-wireframe-schema.json` and `image-wireframe-update-schema.json` files
- Generate the associated TS

### DIFF
--- a/lib/cjs/generated/mobileSessionReplay.d.ts
+++ b/lib/cjs/generated/mobileSessionReplay.d.ts
@@ -41,7 +41,7 @@ export declare type MobileFullSnapshotRecord = CommonRecordSchema & {
 /**
  * Schema of a Wireframe type.
  */
-export declare type Wireframe = ShapeWireframe | TextWireframe;
+export declare type Wireframe = ShapeWireframe | TextWireframe | ImageWireframe;
 /**
  * Schema of all properties of a ShapeWireframe.
  */
@@ -154,6 +154,23 @@ export declare type TextPosition = {
     };
 };
 /**
+ * Schema of all properties of a ImageWireframe.
+ */
+export declare type ImageWireframe = CommonShapeWireframe & {
+    /**
+     * The type of the wireframe.
+     */
+    readonly type: 'image';
+    /**
+     * base64 representation of the image. Can have a null value
+     */
+    base64: string | null;
+    /**
+     * Extension of the image file
+     */
+    imageFileExtension: string;
+};
+/**
  * Mobile-specific. Schema of a Record type which contains mutations of a screen.
  */
 export declare type MobileIncrementalSnapshotRecord = CommonRecordSchema & {
@@ -207,7 +224,7 @@ export declare type MobileMutationPayload = {
 /**
  * Schema of a WireframeUpdateMutation type.
  */
-export declare type WireframeUpdateMutation = TextWireframeUpdate | ShapeWireframeUpdate;
+export declare type WireframeUpdateMutation = TextWireframeUpdate | ShapeWireframeUpdate | ImageWireframeUpdate;
 /**
  * Schema of all properties of a TextWireframeUpdate.
  */
@@ -238,6 +255,23 @@ export declare type ShapeWireframeUpdate = CommonShapeWireframeUpdate & {
      * The type of the wireframe.
      */
     readonly type: 'shape';
+};
+/**
+ * Schema of all properties of a ImageWireframeUpdate.
+ */
+export declare type ImageWireframeUpdate = CommonShapeWireframeUpdate & {
+    /**
+     * The type of the wireframe.
+     */
+    readonly type: 'image';
+    /**
+     * base64 representation of the image. Can have a null value
+     */
+    base64?: string | null;
+    /**
+     * Extension of the image file
+     */
+    imageFileExtension?: string;
 };
 /**
  * Schema of a TouchData.
--- a/lib/cjs/generated/mobileSessionReplay.d.ts
+++ b/lib/cjs/generated/mobileSessionReplay.d.ts
@@ -162,13 +162,13 @@ export declare type ImageWireframe = CommonShapeWireframe & {
      */
     readonly type: 'image';
     /**
-     * base64 representation of the image. Can have a null value
+     * base64 representation of the image. Not required as the ImageWireframe can be initialised without any base64
      */
-    base64: string | null;
+    base64?: string;
     /**
-     * Extension of the image file
+     * MIME type of the image file
      */
-    imageFileExtension: string;
+    mimeType?: string;
 };
 /**
  * Mobile-specific. Schema of a Record type which contains mutations of a screen.
@@ -265,13 +265,13 @@ export declare type ImageWireframeUpdate = CommonShapeWireframeUpdate & {
      */
     readonly type: 'image';
     /**
-     * base64 representation of the image. Can have a null value
+     * base64 representation of the image. Not required as the ImageWireframe can be initialised without any base64
      */
-    base64?: string | null;
+    base64?: string;
     /**
-     * Extension of the image file
+     * MIME type of the image file
      */
-    imageFileExtension?: string;
+    mimeType?: string;
 };
 /**
  * Schema of a TouchData.
--- a/schemas/session-replay/mobile/image-wireframe-schema.json
+++ b/schemas/session-replay/mobile/image-wireframe-schema.json
@@ -9,7 +9,7 @@
       "$ref": "_common-shape-wireframe-schema.json"
     },
     {
-      "required": ["type", "base64", "imageFileExtension"],
+      "required": ["type"],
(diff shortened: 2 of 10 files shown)


schemas/session-replay/mobile/image-wireframe-schema.json
```diff
@@ -0,0 +1,33 @@
...
+          "description": "The type of the wireframe.",
+          "const": "image",
+          "readOnly": true
+        },
+        "base64": {
+          "type": ["string", "null"],
+          "description": "base64 representation of the image. Can have a null value",
+          "readOnly": false
+        },
+        "imageFileExtension": {
```
acorretti: suggestion(blocking): why not sending a `contentType` attribute instead? this is assuming the image has a filename, which may not.

I'd like to hear @mariusc83 thoughts on this.

schemas/session-replay/mobile/image-wireframe-schema.json
```diff
@@ -0,0 +1,33 @@
...
+          "type": "string",
+          "description": "Extension of the image file",
+          "readOnly": false
+        }
+      }
```
acorretti, line 30: suggestion: may we add a unique identifier for the image? maybe optional for now.

lib/cjs/generated/mobileSessionReplay.d.ts
```diff
@@ -153,6 +153,23 @@ export declare type TextPosition = {
...
+     */
+    readonly type: 'image';
+    /**
+     * base64 representation of the image. Can have a null value
+     */
+    base64: string | null;
+    /**
+     * Extension of the image file
+     */
+    imageFileExtension: string;
```
maciejburda: maybe `imageFileFormat` 🤔 (I know that I suggested the current one, but made me think :D)

lib/cjs/generated/mobileSessionReplay.d.ts
```diff
@@ -153,6 +153,23 @@ export declare type TextPosition = {
...
+ */
+export declare type ImageWireframe = CommonShapeWireframe & {
+    /**
+     * The type of the wireframe.
+     */
+    readonly type: 'image';
+    /**
+     * base64 representation of the image. Can have a null value
+     */
+    base64: string | null;
```
maciejburda: maybe `imageDataBase64` 🤔

maciejburda (approved):
Looks great! Tiny suggestion on the naming. Not a blocker though :)

schemas/session-replay/mobile/image-wireframe-schema.json
maciejburda: On iOS we will be always sending PNGs (at least in the first iteration).

acorretti: Okay, I'm thinking mime-types can be more cross-platform and future proof.

schemas/session-replay/mobile/image-wireframe-schema.json
```diff
@@ -0,0 +1,33 @@
...
+      "required": ["type", "base64", "imageFileExtension"],
+      "properties": {
+        "type": {
+          "type": "string",
+          "description": "The type of the wireframe.",
+          "const": "image",
+          "readOnly": true
+        },
+        "base64": {
+          "type": ["string", "null"],
```
ncreated: Is there a different way we could describe this type? I know it won't be the same but making `base64` not `required` will make it optional. Then we could be sending `null` as convention on this field (after all this is temporary measure).

I'm asking because this definition is a new concept to the entire schema. Not sure Android side (@mariusc83 ?), but iOS code generator fails on handling it:
```
🛑 Decoding [CodingKeys(stringValue: "allOf", intValue: nil), _JSONKey(stringValue: "Index 1", intValue: 1), CodingKeys(stringValue: "properties", intValue: nil), _JSONKey(stringValue: "base64", intValue: nil)] is not supported in `JSONSchema.init(from:)`.
→ coding path: allOf → 1 → properties → base64 → type
```
That means, we won't be able to generate the code without first enhancing generator. If it's possible to describe it differently, we will save at least few days iOS side (such type duality is far from Swift idioms).

schemas/session-replay/mobile/image-wireframe-schema.json
ncreated: 👍 for `mimeType: String`, "extension" is vague

schemas/session-replay/mobile/image-wireframe-update-schema.json
```diff
@@ -0,0 +1,33 @@
...
+          "type": ["string", "null"],
+          "description": "base64 representation of the image. Can have a null value",
+          "readOnly": false
+        },
+        "imageFileExtension": {
+          "type": "string",
+          "description": "Extension of the image file",
+          "readOnly": false
+        }
+      }
```
ThibautGeriz: **question:** we can change the image content and/or format? For my it makes little sense

schemas/session-replay/mobile/image-wireframe-schema.json
ThibautGeriz: **suggestion(not-blocking):** I would go the extra mile and type this field with an enum with all the possible types we can expect See list here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

ThibautGeriz:
**suggestion(not-blocking):** add sample that will be use for validation in `samples/session-replay/mobile/record`

schemas/session-replay/mobile/image-wireframe-update-schema.json
maciejburda: The idea is to start with empty image wireframe (just size) and provide base64 after we calculate it on the SDK side.

schemas/session-replay/mobile/image-wireframe-schema.json
mariusc83: Yes I wanted to say the same thing, it will bring some new changes into our generator that I would gladly want to avoid if possible. I agree with @ncreated proposal to have it as an optional.

schemas/session-replay/mobile/image-wireframe-schema.json
mariusc83: @acorretti it already has it as it inherits from `_common-wireframe-schema`

schemas/session-replay/mobile/image-wireframe-schema.json
mquentin: I don't mind going for:
- base64 not required
- base64 type:string

so if you dont have it, it's not there and then you send a mutation that sets it.

I am not fan of the non-required and `"type": ["string", "null"],` as it's not clear how we differentiate the missing field case VS the null case

schemas/session-replay/mobile/image-wireframe-schema.json
mariusc83: +1 for `mimeType` and for @ThibautGeriz suggestion...even though there are lots of extensions there. Just FYI from Android I expect to send `.webp` ;)

schemas/session-replay/mobile/image-wireframe-update-schema.json
mariusc83: Yes but what @ThibautGeriz says makes sense, you will only change the content and not the format (file extension)

schemas/session-replay/mobile/image-wireframe-schema.json
mquentin: For the naming I opened a thread to discuss it. For the type guard of the possible values, I think it a good move but I would do it in a second step not to add more constraints on the SDKs POC implementations (here is the followup jira https://datadoghq.atlassian.net/browse/REPLAY-1210)

schemas/session-replay/mobile/image-wireframe-schema.json
mquentin: I would say the wireframe id would be the unique identifier for now. If we find a way (though the POC implementation) to reuse a base64, I will add this identifier in a follow up PR.

schemas/session-replay/mobile/image-wireframe-schema.json
mquentin: Note: I will go for the same approach for the mimeType as I guess you will set both fields once the base64 is computed

maciejburda (approved):
LGTM

mariusc83 (approved):

ncreated (approved):

BenoitZugmeyer (approved):
Not quite DRY but looks fine :)

mquentin:
Can be merged: failing https://github.com/DataDog/rum-events-format/actions/runs/3985208980/jobs/6832308340 is flaky (doubled check with RUM sdk)

PR #128, opened 2023-01-30
[RUMF-1450] Browser Mouse Interaction: remove x/y for blur and focus

bcaudan:
# Motivation

Since  on focus and blur events x and y are not used by the player, browser-sdk will stop to compute them to avoid an issue on safari and useless computation.

# Changes

Split mouse interaction schema in two, with or without x/y depending on type.

### DIFF
--- a/schemas/session-replay/browser/mouse-interaction-schema.json
+++ b/schemas/session-replay/browser/mouse-interaction-schema.json
@@ -4,25 +4,44 @@
   "title": "MouseInteraction",
   "type": "object",
   "description": "Browser-specific. Schema of a MouseInteraction.",
-  "required": ["type", "id", "x", "y"],
-  "properties": {
-    "type": {
-      "type": "integer",
-      "enum": [0, 1, 2, 3, 4, 5, 6, 7, 9],
-      "description": "The type of MouseInteraction.",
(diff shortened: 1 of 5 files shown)


schemas/session-replay/browser/mouse-interaction-schema.json
```diff
@@ -4,25 +4,44 @@
...
-      "enum": [0, 1, 2, 3, 4, 5, 6, 7, 9],
-      "description": "The type of MouseInteraction.",
-      "readOnly": true
+  "oneOf": [
+    {
+      "required": ["type", "id", "x", "y"],
+      "properties": {
+        "type": {
+          "type": "integer",
+          "enum": [0, 1, 2, 3, 4, 7, 9],
```
xgouchet, line 13: Maybe it could be helpful to have in the description the meaning for each value. Apparently 5 and 6 means blur and focus, but what about the others?

BenoitZugmeyer (approved):

schemas/session-replay/browser/mouse-interaction-schema.json
bcaudan: added to the description

PR #135, opened 2023-03-28
[RUMF-1489] Add view._dd.page_states attribute

amortemousque:
Modern browsers today will sometimes suspend pages or discard them entirely when system resources are constrained, which can lead to unexpected behaviors. The [Page Lifecycle API](https://developer.chrome.com/blog/page-lifecycle-api/), provides lifecycle hooks so we can detect and handle these browser interventions.

This PR, collect page lifecycle state changes during a view to help Browser SDK issue investigations.

### DIFF
--- a/schemas/rum/view-schema.json
+++ b/schemas/rum/view-schema.json
@@ -329,7 +329,7 @@
                   },
                   "start": {
                     "type": "integer",
-                    "description": "Duration in ns to the start of the page state",
+                    "description": "Duration in ns between start of the view and start of the page state",
                     "minimum": 0,
                     "readOnly": true
                   }
--- a/schemas/rum/view-schema.json
+++ b/schemas/rum/view-schema.json
@@ -313,6 +313,30 @@
               "description": "Version of the update of the view event",
               "minimum": 0,
               "readOnly": true
+            },
+            "page_states": {
+              "type": "array",
+              "description": "List of the page states during the view",
+              "items": {
+                "type": "object",
+                "description": "Properties of the page state",
+                "required": ["state", "start"],
+                "properties": {
+                  "state": {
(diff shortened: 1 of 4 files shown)


BenoitZugmeyer (approved):

schemas/rum/view-schema.json
```diff
@@ -313,6 +313,30 @@
...
+                "required": ["state", "start"],
+                "properties": {
+                  "state": {
+                    "enum": ["active", "passive", "hidden", "frozen", "terminated"],
+                    "description": "Page state name",
+                    "readOnly": true
+                  },
+                  "start": {
+                    "type": "integer",
+                    "description": "Duration in ns to the start of the page state",
```
xgouchet: Just clarify the doc and match the `in_foreground_periods.start` attribute.
```suggestion
                    "description": "Duration in ns between start of the view and start of the page state",
```

schemas/rum/view-schema.json
```diff
@@ -313,6 +313,30 @@
...
               "minimum": 0,
               "readOnly": true
+            },
+            "page_states": {
+              "type": "array",
+              "description": "List of the page states during the view",
+              "items": {
+                "type": "object",
+                "description": "Properties of the page state",
+                "required": ["state", "start"],
```
xgouchet, line 323: For the `in_foreground_periods` attributes, we also keep a `duration` attribute, would it make sense to add it to the `page_states`?

amortemousque: For browser SDK investigation it's not needed. Plus, we can still calculate the duration like so: 
`page_states[n + 1].start - page_states[n].start `

xgouchet (approved):

ThibautGeriz (approved):

xgouchet (approved):

kbogtob (approved):

PR #136, opened 2023-04-05
Add start_reason to session object

plousada:
Adds `start_reason` field to sessions in order to bring clarity on the motives that lead to new sessions being created for a given app. 
The main motivation is to have this information easily queryable for support requests and customer questions on the amount of sessions they are seeing / being billed for.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -680,6 +680,16 @@ export declare type RumViewEvent = CommonProperties & {
         [k: string]: unknown;
     };
     /**
+     * Session properties
+     */
+    readonly session?: {
+        /**
+         * The precondition that led to the creation of the session
+         */
+        readonly start_reason?: 'app_start' | 'inactivity_timeout' | 'max_duration' | 'stop_api' | 'background_event';
+        [k: string]: unknown;
+    };
+    /**
      * Feature flags properties
      */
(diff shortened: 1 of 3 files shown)


schemas/rum/_common-schema.json
```diff
@@ -63,6 +63,12 @@
           "description": "Whether this session is currently active. Set to false to manually stop a session",
           "default": true,
           "readOnly": true
+        },
+        "start_reason": {
+          "type": "string",
+          "description": "The precondition that led to the creation of the session",
+          "enum": ["app_start", "15m_expire", "4h_expire", "stop_api", "background_event"],
```
0xnm: I would say it is better to use something like `inactivity_timeout` and `max_duration` instead of `15m_expire` and `4h_expire`, because essentially these constants may change at some point

xgouchet (approved):

0xnm (approved):

xgouchet (approved):

BenoitZugmeyer (approved):

0xnm (approved):

xgouchet (approved):

bcaudan (approved):

BenoitZugmeyer (approved):
Nice to move it to RumViewEvent! We should do the same for `session.is_active`, I'll create a PR

louisj-dd (approved):

PR #140, opened 2023-04-20
Add scroll metrics to the view schema

hamza-kadiri:
Add new scroll metrics to the view event schema

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -736,19 +736,19 @@ export declare type RumViewEvent = CommonProperties & {
          */
         readonly scroll?: {
             /**
-             * Maximum scrolled distance on this view (in pixels)
+             * Distance between the top and the lowest point reached on this view (in pixels)
              */
             readonly max_depth?: number;
             /**
-             * Page scroll height when the maximum scrolled distance was reached for this view (in pixels)
+             * Page scroll height (total height) when the maximum scroll depth was reached for this view (in pixels)
              */
             readonly max_depth_scroll_height?: number;
             /**
-             * Page scroll top when the maximum scrolled distance was reached for this view (in pixels)
+             * Page scroll top (scrolled distance) when the maximum scroll depth was reached for this view (in pixels)
              */
             readonly max_depth_scroll_top?: number;
             /**
-             * Timestamp of the scroll event that reached the maximum scroll depth for this view
+             * Duration between the view start and the scroll event that reached the maximum scroll depth for this view (in nanoseconds)
              */
             readonly max_depth_time?: number;
             [k: string]: unknown;
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -738,19 +738,19 @@ export declare type RumViewEvent = CommonProperties & {
             /**
              * Maximum scrolled distance on this view (in pixels)
              */
-            readonly max_depth?: number;
+            readonly max_depth: number;
             /**
              * Maximum scrollable distance on this view (in pixels)
              */
-            readonly max_scroll_height?: number;
+            readonly max_scroll_height: number;
             /**
              * Timestamp of the scroll event that reached the maximum scroll depth for this view
              */
-            readonly max_depth_time?: number;
+            readonly max_depth_time: number;
             /**
              * Maximum amount of pixels scrolled
              */
-            readonly max_scroll_top?: number;
+            readonly max_scroll_top: number;
             [k: string]: unknown;
         };
         [k: string]: unknown;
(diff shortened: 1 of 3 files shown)


schemas/rum/view-schema.json
```diff
@@ -360,6 +360,39 @@
...
+                },
+                "max_depth_time": {
+                  "type": "number",
+                  "description": "Timestamp of the scroll event that reached the maximum scroll depth for this view",
+                  "readOnly": true
+                },
+                "max_scroll_top": {
+                  "type": "number",
+                  "description": "Maximum amount of pixels scrolled",
+                  "readOnly": true
```
amortemousque: #

**❓ question:** ‏Should these properties be required?

schemas/rum/view-schema.json
```diff
@@ -360,6 +360,40 @@
...
+                  "type": "number",
+                  "description": "Maximum scrollable distance on this view (in pixels)",
+                  "readOnly": true
+                },
+                "max_depth_time": {
+                  "type": "number",
+                  "description": "Timestamp of the scroll event that reached the maximum scroll depth for this view",
+                  "readOnly": true
+                },
+                "max_scroll_top": {
```
xgouchet: What's the difference between this and the `max_depth`?

hamza-kadiri: `max_depth` is a measurement of "how far did a user scroll on the page", this is the main metric we're interested in for scroll maps.  The `scroll_top` is the amount of pixels the user scrolled. We basically have: `depth = viewport_height + scroll_top`. We also need it to compute the number of views/sessions where the user scrolled at least once (`@display.scroll.max_scroll_top:>0`)

So the difference between the two is the viewport height.

schemas/rum/view-schema.json
```diff
@@ -360,6 +360,40 @@
             }
           },
           "readOnly": true
+        },
+        "display": {
+          "type": "object",
+          "description": "Display properties",
```
bcaudan, line 395: **💬 suggestion:** ‏`display` group should be read only as well

schemas/rum/view-schema.json
```diff
@@ -360,6 +360,40 @@
...
+                  "readOnly": true
+                },
+                "max_scroll_height": {
+                  "type": "number",
+                  "description": "Maximum scrollable distance on this view (in pixels)",
+                  "readOnly": true
+                },
+                "max_depth_time": {
+                  "type": "number",
+                  "description": "Timestamp of the scroll event that reached the maximum scroll depth for this view",
```
bcaudan: **💭 thought:** ‏~~not sure how this timestamp will be used but event dates are corrected by the backend, will this one be corrected as well? if not, could it be bothering?~~
Looking at the implementation, it seems that it is not a timestamp but a duration. If that is it, it could be interesting to specify what is the reference for this duration (view start?).

schemas/rum/view-schema.json
xgouchet: Oh ok makes sense

xgouchet (approved):

xgouchet (approved):

schemas/rum/view-schema.json
```diff
@@ -360,6 +360,40 @@
...
+                  "readOnly": true
+                },
+                "max_depth_scroll_top": {
+                  "type": "number",
+                  "description": "Page scroll top when the maximum scrolled distance was reached for this view (in pixels)",
+                  "readOnly": true
+                },
+                "max_depth_time": {
+                  "type": "number",
+                  "description": "Timestamp of the scroll event that reached the maximum scroll depth for this view",
```
bcaudan: is it a timestamp or a duration in ns?

bcaudan: **🥜 nitpick:** ‏for consistency with other descriptions
```suggestion
                "max_depth_time": {
                  "type": "number",
                  "description": "Timestamp of the scroll event that reached the maximum scroll distance for this view",
```

lib/cjs/generated/rum.d.ts
```diff
@@ -749,6 +749,34 @@ export declare type RumViewEvent = CommonProperties & {
...
+             */
+            readonly max_depth_scroll_height?: number;
+            /**
+             * Page scroll top (scrolled distance) when the maximum scroll depth was reached for this view (in pixels)
+             */
+            readonly max_depth_scroll_top?: number;
+            /**
+             * Duration between the view start and the scroll event that reached the maximum scroll depth for this view (in nanoseconds)
+             */
+            readonly max_depth_time?: number;
```
amortemousque: Shouldn't these attribute be required?

hamza-kadiri: oops, I forgot to add them back, thanks for catching

amortemousque (approved):

bcaudan (approved):

detzgk (approved):

PR #145, opened 2023-06-02
RUMF-1592: Send the views session replay default privacy level

acorretti:
Addition in order to enable reporting Session Replay's `defaultPrivacyLevel` init configuration property on each view.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -704,6 +704,16 @@ export declare type RumViewEvent = CommonProperties & {
         [k: string]: unknown;
     };
     /**
+     * Session Replay properties
+     */
+    replay?: {
+        /**
+         * Default Session Replay privacy level established for this view
+         */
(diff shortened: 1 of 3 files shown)


lib/cjs/generated/rum.d.ts
```diff
@@ -703,6 +703,16 @@ export declare type RumViewEvent = CommonProperties & {
...
         [k: string]: unknown;
     };
+    /**
+     * Session Replay properties
+     */
+    replay?: {
+        /**
+         * Default Session Replay privacy level established for this view
+         */
+        readonly privacyLevel?: 'mask' | 'mask-user-input' | 'allow';
```
Manu-Tran, line 713: Shouldn't we use snake_case instead of camelCase to match the other fields ?

Manu-Tran (commented):
Small NIT

lib/cjs/generated/rum.d.ts
acorretti: Absolutely! But there'll be a new PR, @amortemousque 's RFC is renaming this field

PR #148, opened 2023-06-12
[RUMF-1597] updates for browser billing changes

bcaudan:
# Motivation 

Since billing will be based on `has_replay` field, SDKs could stop sending the `_dd.session.plan` with all events.
Sampling result can still be interesting for customers if they want to know why a session wasn’t recorded.

# Changes

- deprecate `_dd.session.plan`
- add `session.sampled_for_replay` on view event

### DIFF
--- a/schemas/rum/view-schema.json
+++ b/schemas/rum/view-schema.json
@@ -313,6 +313,11 @@
               "description": "Whether this session is currently active. Set to false to manually stop a session",
               "default": true,
               "readOnly": true
+            },
+            "sampled_for_replay": {
+              "type": "boolean",
+              "description": "Whether this session has been sampled for replay",
+              "readOnly": true
             }
           },
           "readOnly": true
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -954,9 +954,9 @@ export interface CommonProperties {
          */
         session?: {
             /**
-             * Session plan: 1 is the plan without replay, 2 is the plan with replay
+             * Session plan: 1 is the plan without replay, 2 is the plan with replay (deprecated)
              */
-            plan: 1 | 2;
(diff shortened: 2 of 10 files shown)


xgouchet (approved):

schemas/rum/view-schema.json
```diff
@@ -313,6 +313,11 @@
               "description": "Whether this session is currently active. Set to false to manually stop a session",
               "default": true,
               "readOnly": true
+            },
+            "sampled_for_replay": {
+              "type": "boolean",
```
acorretti, line 318: 🥜 (nitpick) I'm fine with this or either an `"enum": [ true ]` if this value will always be true when sent

acorretti (approved):

schemas/rum/view-schema.json
bcaudan: I had in mind to also send `false` values, it seems helpful to have `true` and `false` values in the events.
Any reason to not send them?

acorretti: It was either that model, or the `has_replay` one. Works either way!

mariusc83 (approved):
LGTM

detzgk (approved):
LGTM for the schema changes.

On the backend side, we currently default to the legacy plan if the parameter is unset. Should we be worried about this or will you continue sending this parameter until the backend changes are made?

bcaudan:
> LGTM for the schema changes.
> 
> On the backend side, we currently default to the legacy plan if the parameter is unset. Should we be worried about this or will you continue sending this parameter until the backend changes are made?

The plan is to stop sending it with the next major release of the browser SDK, so will not start to send it before backend changes are made 🙂
I planned to ping you guys to sync about it.

kbogtob (approved):

PR #153, opened 2023-07-05
[RUM-235] add configured sample rates fields to all RUM events

BenoitZugmeyer:
This PR adds the following fields to all RUM events:

* `@_dd.configuration.session_sample_rate`
* `@_dd.configuration.session_replay_sample_rate`

Those fields are private and should not be used by customers. They will reflect the corresponding initialization parameter values.

### DIFF
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -300,6 +300,27 @@
             }
           }
         },
+        "configuration": {
+          "type": "object",
+          "description": "Subset of the configuration options used to initialize the SDK",
+          "readOnly": true,
+          "properties": {
+            "session_sample_rate": {
+              "type": "integer",
+              "description": "The percentage of sessions tracked",
+              "minimum": 0,
+              "maximum": 100,
+              "readOnly": true
+            },
+            "session_replay_sample_rate": {
+              "type": "integer",
+              "description": "The percentage of sessions with Browser RUM & Session Replay pricing tracked",
+              "minimum": 0,
+              "maximum": 100,
+              "readOnly": true
+            }
+          }
+        },
         "browser_sdk_version": {
           "type": "string",
           "description": "Browser SDK version",
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -307,14 +307,14 @@
           "required": ["session_sample_rate", "session_replay_sample_rate"],
           "properties": {
             "session_sample_rate": {
-              "type": "integer",
+              "type": "number",
               "description": "The percentage of sessions tracked",
               "minimum": 0,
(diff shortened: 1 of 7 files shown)


schemas/rum/_common-schema.json
```diff
@@ -300,6 +300,27 @@
...
+            "session_sample_rate": {
+              "type": "integer",
+              "description": "The percentage of sessions tracked",
+              "minimum": 0,
+              "maximum": 100,
+              "readOnly": true
+            },
+            "session_replay_sample_rate": {
+              "type": "integer",
+              "description": "The percentage of sessions with Browser RUM & Session Replay pricing tracked",
```
xgouchet: Mobile has session replay too

```suggestion
              "description": "The percentage of sessions with RUM & Session Replay pricing tracked",
```

schemas/rum/_common-schema.json
```diff
@@ -300,6 +300,27 @@
             }
           }
         },
+        "configuration": {
+          "type": "object",
+          "description": "Subset of the configuration options used to initialize the SDK",
+          "readOnly": true,
```
bcaudan: **💬 suggestion:** 

- description: FMU, we should send a value even if the customer use the default option
- required fields: if configuration is present, we should always have session sample rate, not sure for session replay rate on mobile, it could worth double checking with them

‏
```suggestion
        "configuration": {
          "type": "object",
          "description": "Subset of the configuration options used by the SDK",
          "required": ["session_sample_rate"]
          "readOnly": true,
```

BenoitZugmeyer: I reworded the description, let me know what you think.
I made both fields as required since there is a session replay sample rate for mobile too.

schemas/rum/_common-schema.json
```diff
@@ -300,6 +300,28 @@
...
+          "properties": {
+            "session_sample_rate": {
+              "type": "integer",
+              "description": "The percentage of sessions tracked",
+              "minimum": 0,
+              "maximum": 100,
+              "readOnly": true
+            },
+            "session_replay_sample_rate": {
+              "type": "integer",
```
bcaudan: **💬 suggestion:** ‏it seems that it should be number to accept floating point numbers (cf [doc](https://json-schema.org/understanding-json-schema/reference/numeric.html)). 
What about adding a test for this one?
```suggestion
              "type": "number",
              "description": "The percentage of sessions tracked",
              "minimum": 0,
              "maximum": 100,
              "readOnly": true
            },
            "session_replay_sample_rate": {
              "type": "number",
```

bcaudan (approved):

xgouchet (approved):

Manu-Tran (approved):

PR #159, opened 2023-08-23
✨ [RUM-265] add telemetry configuration `store_contexts_across_pages`

bcaudan:
Add telemetry configuration for new configuration to store contexts in local storage.

### DIFF
--- a/schemas/telemetry/configuration-schema.json
+++ b/schemas/telemetry/configuration-schema.json
@@ -112,6 +112,10 @@
                   "type": "boolean",
                   "description": "Whether it is allowed to use LocalStorage when cookies are not available"
                 },
+                "store_contexts_across_pages": {
+                  "type": "boolean",
+                  "description": "Whether contexts are stored in local storage"
+                },
                 "allow_untrusted_events": {
                   "type": "boolean",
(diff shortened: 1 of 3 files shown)


schemas/telemetry/configuration-schema.json
```diff
@@ -112,6 +112,10 @@
                   "type": "boolean",
                   "description": "Whether it is allowed to use LocalStorage when cookies are not available"
                 },
+                "store_contexts_across_pages": {
+                  "type": "boolean",
+                  "description": "Whether contexts are stored in local storage"
+                },
```
xgouchet, line 118: The description and name don't really say the same thing. Does the `store_contexts_across_pages` feature necessarily means using local storage?

xgouchet (approved):

schemas/telemetry/configuration-schema.json
bcaudan: Indeed, my intent was to add a bit more details than what is already conveyed by the name of the configuration options.
"Across pages" is more user facing and "in local storage" is more from a technical point of view.
We only have local storage for now, no worries to match name and description if it is disturbing.

amortemousque (approved):

PR #162, opened 2023-09-08
[RUM-311] Add graphql property to resources

louiszawadzki:
This PR adds a `"graphql"` property for better supporting resources calling a graphql endpoint.

The following properties are added in `"graphql"`:
- `operationType`: the type of GraphQL operation (can be `query`, `mutation` or `subscription`)
  - `subscription` will not be reported by SDKs and frontend in the first iteration
- `payload` (optional): the actual content of the operation as a string
  - optional as it might actually contain a lot of data
  - `readonly: false` as it could potentially contain sensitive information (if used in an improper way) so we should enable users to overwrite it 
- `operationName` (optional): the name of the operation if any
  - optional as developers have the choice to name of the GraphQL operation or not
- `variables` (optional): the variables passed to the graphql operation
  - optional as they might not be defined
  - `readonly: false` as they could potentially contain sensitive information so we should enable users to overwrite it 

Link to full internal RFC: https://datadoghq.atlassian.net/l/cp/iyo7Bg1P

### DIFF
--- a/schemas/rum/resource-schema.json
+++ b/schemas/rum/resource-schema.json
@@ -223,6 +223,30 @@
                 }
               },
               "readOnly": true
+            },
+            "graphql": {
+              "type": "object",
+              "description": "GraphQL requests parameters",
+              "required": ["operationType", "query"],
+              "properties": {
+                "operationType": {
+                  "type": "string",
+                  "description": "Type of the GraphQL operation",
+                  "enum": ["query", "mutation", "subscription"],
+                  "readOnly": true
+                },
+                "query": {
+                  "type": "string",
+                  "description": "Content of the GraphQL operation",
+                  "readOnly": false
+                },
+                "operationName": {
+                  "type": "string",
+                  "description": "Name of the GraphQL operation",
+                  "readOnly": true
+                }
+              },
+              "readOnly": true
             }
           },
           "readOnly": true
--- a/schemas/rum/resource-schema.json
+++ b/schemas/rum/resource-schema.json
@@ -227,7 +227,7 @@
             "graphql": {
               "type": "object",
               "description": "GraphQL requests parameters",
-              "required": ["operationType", "query"],
+              "required": ["operationType"],
               "properties": {
                 "operationType": {
                   "type": "string",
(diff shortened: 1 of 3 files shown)


schemas/rum/resource-schema.json
```diff
@@ -223,6 +223,30 @@
...
+            },
+            "graphql":{
+              "type": "object",
+              "description": "GraphQL requests parameters",
+              "required": ["operationType", "query"],
+              "properties": {
+                "operationType": {
+                  "type": "string",
+                        "description": "Type of the GraphQL operation",
+                  "enum": ["query", "mutation", "subscription"],
```
bcaudan: weird formatting 🤔
It seems that the ci was not running anymore, I updated the configuration so now it should raise an issue 🙂

schemas/rum/resource-schema.json
```diff
@@ -223,6 +223,30 @@
...
+              "description": "GraphQL requests parameters",
+              "required": ["operationType", "query"],
+              "properties": {
+                "operationType": {
+                  "type": "string",
+                  "description": "Type of the GraphQL operation",
+                  "enum": ["query", "mutation", "subscription"],
+                  "readOnly": true
+                },
+                "query": {
```
xgouchet: In the context of Graphql, this sounds like this could be ambiguous (e.g. customers wondering why there's a `query` attribute for a mutation `request`). Maybe we could use something more generic, like `operationContent` or `payload`. 
Also, this is marked as mandatory attribute, meaning that we're essentially duplicating the upload usage for graphql request (payload is sent once in the graphql request, and another time in datadog). For queries, payload can be quite small but mutation could grow, aren't we afraid of the impact this can have on the end user's data plan?

schemas/rum/resource-schema.json
louiszawadzki: Indeed, that's been updated!

schemas/rum/resource-schema.json
louiszawadzki: I agree on the `query` naming, I will look for the exact term in the GraphQL docs.

louiszawadzki: I also tend to think it should not be mandatory so that we can at least have the opportunity to drop entirely the content.
I will start an internal conversation to decide on whether to report the payload or not.

fuzzybinary: The Flutter GraphQL client has an issue where it's difficult to get the full payload (because it's already parsed by the time it hits the gql_link). I would appreciate it if the payload was optional.

schemas/rum/resource-schema.json
```diff
@@ -223,6 +223,35 @@
...
+                  "description": "Type of the GraphQL operation",
+                  "enum": ["query", "mutation", "subscription"],
+                  "readOnly": true
+                },
+                "operationName": {
+                  "type": "string",
+                  "description": "Name of the GraphQL operation",
+                  "readOnly": true
+                },
+                "payload": {
```
xgouchet, line 243: Much better 👍

xgouchet (approved):

kbogtob (approved):

bcaudan (approved):

PR #164, opened 2023-10-11
RUM-1507 Add image resource identifier

maciejburda:
Adds a new field that will be utilised in [Resource Endpoint](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3224240335/Replay+Resource+Service) work.

It also brings [significant optimisation to Diffing](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3189606163/RFC+-+Reducing+CPU+Pressure+-+iOS+Session+Replay) algorithms in mobile SDKs.

### DIFF
--- a/lib/esm/generated/sessionReplay.d.ts
+++ b/lib/esm/generated/sessionReplay.d.ts
@@ -467,6 +467,10 @@ export declare type ImageWireframe = CommonShapeWireframe & {
      */
     base64?: string;
     /**
+     * Unique identifier of the image resource
+     */
+    resourceIdentifier?: string;
+    /**
      * MIME type of the image file
      */
     mimeType?: string;
@@ -586,6 +590,10 @@ export declare type ImageWireframeUpdate = CommonShapeWireframeUpdate & {
      * base64 representation of the image. Not required as the ImageWireframe can be initialised without any base64
      */
     base64?: string;
+    /**
+     * Unique identifier of the image resource
+     */
+    resourceIdentifier?: string;
     /**
      * MIME type of the image file
      */
--- a/lib/esm/generated/sessionReplay.d.ts
+++ b/lib/esm/generated/sessionReplay.d.ts
@@ -469,7 +469,7 @@ export declare type ImageWireframe = CommonShapeWireframe & {
     /**
      * Unique identifier of the image resource
      */
-    resourceIdentifier?: string;
+    resourceId?: string;
     /**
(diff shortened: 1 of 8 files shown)


lib/esm/generated/sessionReplay.d.ts
```diff
@@ -586,6 +590,10 @@ export declare type ImageWireframeUpdate = CommonShapeWireframeUpdate & {
      * base64 representation of the image. Not required as the ImageWireframe can be initialised without any base64
      */
     base64?: string;
+    /**
+     * Unique identifier of the image resource
+     */
+    resourceIdentifier?: string;
```
mariusc83: hmmm...should we do it optional ?

maciejburda: Not a TS expert, but I think it already is (hence the `?` at the end of the name)

within JSON there's array: ```"required": ["type"],``` which I haven't touched for this one

mariusc83: the way it should be : `resourceIdentifier: string` and then if you do not want it required (you want it optional) you do not add it in the `required` fields. But I think this field should not be optional. @mquentin what do you think ?

ThibautGeriz: > Not a TS expert, but I think it already is (hence the ? at the end of the name)

yes you are correct.

Normally, you generated it with the CLI and it's set as optional because it's not in `required`

schemas/session-replay/mobile/image-wireframe-update-schema.json
```diff
@@ -22,6 +22,11 @@
           "description": "base64 representation of the image. Not required as the ImageWireframe can be initialised without any base64",
           "readOnly": false
         },
+        "resourceIdentifier": {
```
ThibautGeriz: **question:** why not simply call it `id`?

maciejburda: `id` is already taken by the wireframe itself.

I wanted to be descriptive with the property name, but I guess `resourceId` would follow the convention more.

WDYT?

ThibautGeriz: sound good

ThibautGeriz (approved):

mariusc83 (approved):

mquentin (approved):

PR #165, opened 2023-10-12
[RUM][Replay][Webviews] introduce webview wireframe and a nestedEnvId for record generated by a webview

mquentin:
### What

As part of the webview support efforts, records need to be tagged if they were produced from an instrumented webview so we can stream them to the right encapsulation replay environment.
The record that are supported by webviews will then have an optional field `nestedEnvId` that describe the id of the webview that generated the records.
As in the future, we could also support the replaying of other encapsulated environments such as Iframes, it's safer to leverage an agnostic `nestedEnvId` rather than a `webViewId`

On top of that, the new `WebviewWireframe` schema need to be created. It inherits all the fields of a `ShapeWireframe` with an added `nestedEnvId` that describe the unique replayed webview environment that will be encapsulated in this container.

### Then

All the browser specific records as well as the common records such as `MetaRecord`, `FocusRecord`, `VisualViewportRecord`... that are supported by webviews inherit the new and optional `nestedEnvId` field.

The `webview-wireframe-schema.json` and its update schemas are defined

### DIFF
--- a/schemas/session-replay/common/_webview-supported-common-record-schema.json
+++ b/schemas/session-replay/common/_webview-supported-common-record-schema.json
@@ -1,0 +1,21 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/common/_webview-supported-common-record-schema.json",
+  "title": "WebviewSupportedCommonRecordSchema",
+  "type": "object",
+  "description": "Schema of common properties for a Record event type that is supported by webviews.",
+  "allOf": [
+    {
+      "$ref": "_common-record-schema.json"
+    },
+    {
+      "properties": {
+        "nestedEnvId": {
+          "type": "integer",
+          "description": "Defines the unique ID of the nested replay environment that generated this record.",
+          "readOnly": true
+        }
+      }
+    }
+  ]
+}
--- a/schemas/session-replay/common/view-end-record-schema.json
+++ b/schemas/session-replay/common/view-end-record-schema.json
@@ -6,7 +6,7 @@
   "description": "Schema of a Record which signifies that view lifecycle ended.",
   "allOf": [
     {
-      "$ref": "_common-record-schema.json"
+      "$ref": "_webview-supported-common-record-schema.json"
     },
     {
       "required": ["type"],
--- a/schemas/session-replay/mobile/webview-wireframe-update-schema.json
+++ b/schemas/session-replay/mobile/webview-wireframe-update-schema.json
@@ -1,0 +1,28 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/mobile/webview-wireframe-update-schema.json",
+  "title": "WebviewWireframeUpdate",
+  "type": "object",
+  "description": "Schema of all properties of a WebviewWireframeUpdate.",
+  "allOf": [
+    {
+      "$ref": "_common-shape-wireframe-update-schema.json"
+    },
+    {
+      "required": ["type", "nestedEnvId"],
+      "properties": {
+        "type": {
+          "type": "string",
+          "description": "The type of the wireframe.",
+          "const": "webview",
+          "readOnly": true
+        },
+        "nestedEnvId": {
+          "type": "integer",
+          "description": "Defines the unique ID of the replayed webview environment that will be nested in this container.",
+          "readOnly": true
+        }
+      }
+    }
+  ]
+}
--- a/lib/cjs/generated/browserSessionReplay.d.ts
+++ b/lib/cjs/generated/browserSessionReplay.d.ts
@@ -31,12 +31,21 @@ export declare type BrowserRecord = BrowserFullSnapshotRecord | BrowserIncrement
 /**
  * Browser-specific. Schema of a Record type which contains the full snapshot of a screen.
  */
-export declare type BrowserFullSnapshotRecord = CommonRecordSchema & {
+export declare type BrowserFullSnapshotRecord = WebviewSupportedCommonRecordSchema & {
     /**
      * The type of this Record.
      */
     readonly type: 2;
     data: BrowserNode;
+};
+/**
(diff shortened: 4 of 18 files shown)


schemas/session-replay/mobile/webview-wireframe-update-schema.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "required": ["type", "webViewId"],
+      "properties": {
+        "type": {
+          "type": "string",
+          "description": "The type of the wireframe.",
+          "const": "webview",
+          "readOnly": true
+        },
+        "webViewId": {
```
mquentin: this field is read only so it should not be updated. However, even though the wireframe.id would be sufficient to identify the shape to update, the `webViewId` will help up identifying webview container updates out of all the shape wireframe updates.

schemas/session-replay/common/_webview-supported-common-record-schema.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/common/_webview-supported-common-record-schema.json",
+  "title": "WebviewSupportedCommonRecordSchema",
```
mquentin, line 4: This new addition will prevent Mobile FS and Incremental S records to have the possibility to contain a `nestedEnvId`. Indeed such records are generated by the native apps and are not supported by child webviews.

schemas/session-replay/common/view-end-record-schema.json
```diff
@@ -6,7 +6,7 @@
   "description": "Schema of a Record which signifies that view lifecycle ended.",
   "allOf": [
     {
-      "$ref": "_common-record-schema.json"
+      "$ref": "_webview-supported-common-record-schema.json"
```
mariusc83, line 9: hmmmm why do we want to alter all the other records by inheriting this `webview_supported` schema ?

mariusc83: the way I see it this will modify also the mobile records format ?

mquentin: On the first hand, we want to alter only the records that could have a `nestedEnvId` field so far which are the:browser records. The Mobile FS and Incremental S records would never have such field as they are Native replays. As a result, I think it's better not to even give the possibility in the schemas.

On the other, for the shared records like the Metadata that is both generated from the Native Mobile Replay recorder and the Browser replay recorder, we have to enable this optional field.
Indeed, you don't want to apply the Metarecord of the webview on the main Native replay but want to apply it inside the webview encapsulation: we therefore need the `nestedEnvId`.

Another approach would be to split the Metarecord between the BrowserMetarecord (that could have the `nestedEnvId`) and MobileMetarecord (that could not have it) but it would require introducing duplication and new record types.

Maybe a refacto could be planned in the future to clearly separate Mobile and Browser records but I think it is out of the scope of the webview project.

mariusc83: hmmm, I guess we can have this optional field that we do not use in Mobile. Even though right now it is a ongoing discussion on our end as there are plenty of fields there that are coming from browser in our rum events that we do not use and want to go with the splitting strategy as we have in SR. Maybe it is worth considering this as with this change we tend to go back to the old routine we had in rum. @acorretti , @bcaudan what would be the effort to have these schemas separated also like we do with the other records schema ?

mquentin: From my perspective, the concept of Meta record is not Mobile or Browser related. It's Replay related. On top of that, for the replay player it has the exact same goal for mobile and browser so I don't think we should split it.

For specific records like Mobile FS and Mobile incrementals, you do not have the optional field as you still inherit from the `_common-record-schema.json` so it sounds all good isn't it ?

mariusc83 (dismissed):

mariusc83 (dismissed):

schemas/session-replay/mobile/webview-wireframe-update-schema.json
amortemousque: Talking with @bcaudan , he suggested `bridgeId` as naming. This way we are agnostic from whether it's a webview or an iFrame(in the future). I also think it's a nice option. Wdyt?

schemas/session-replay/common/_webview-supported-common-record-schema.json
amortemousque: Talking with @bcaudan, he suggested `bridgeId` as naming. This way we are agnostic from whether it’s a webview or an iFrame(in the future) and we reuse the bridge concept we already have. I also think it’s a nice option. Wdyt?

mquentin: To me SDKs wise it’s good as we would really map the concept of 1 bridge 1 id and better describe the way the webviews are intrumented.

However in the player that has 0 SDK bridge notion, rendering stuff based on a `bridgeId` will be difficult to understand.

But honestly is a weak no for me so if you all understand the format better with `bridgeId` I am good with it

bcaudan: > However in the player that has 0 SDK bridge notion, rendering stuff based on a bridgeId will be difficult to understand.

Indeed.

> As in the future, we could also support the replaying of other encapsulated environments such as Iframes it's safer to leverage an agnostic encapsulatedReplayId rather than a webViewId

Not super fan of `encapsulatedReplayId`:

- encapsulated is a bit of a mouthful, maybe nested or embedded could work too
- replay would be player oriented, this structure is already used by other product. 

Looking at your description, maybe something like `nestedEnvId` could work.

mquentin: `nestedEnvId` sounds great to me ! I will go for it then.

amortemousque (approved):

mariusc83 (approved):

hugolevyy (approved):

PR #167, opened 2023-11-03
[RUM][REPLAY] Add resource schema

ThibautGeriz:
Add schema for resource metadata. We're uploading the resource content: CSS, Image... to the backend.

### DIFF
--- a/schemas/resource-metadata-schema.json
+++ b/schemas/resource-metadata-schema.json
@@ -1,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "resource-metadata-schema.json",
+  "title": "ResourceMetadata",
+  "type": "object",
+  "description": "Schema of the resource metadata. The metadata to idenfify the asset's content uploaded to th backend",
+  "required": ["type", "application"],
+  "properties": {
+    "type": {
+      "const": "resource"
+    },
+    "application": {
+      "type": "object",
+      "description": "Application properties",
+      "required": ["id"],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "UUID of the application",
+          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+          "readOnly": true
+        }
+      },
+      "readOnly": true
+    }
+  }
+}
--- a/schemas/resource-metadata-schema.json
+++ b/schemas/resource-metadata-schema.json
@@ -3,7 +3,7 @@
   "$id": "resource-metadata-schema.json",
   "title": "ResourceMetadata",
   "type": "object",
-  "description": "Schema of the resource metadata. The metadata to idenfify the asset's content uploaded to th backend",
+  "description": "Schema of the resource metadata. The metadata to idenfify the asset's content uploaded to the backend",
   "required": ["type", "application"],
   "properties": {
     "type": {


maciejburda (dismissed):

bcaudan:
I am guessing we are not talking about [RUM resources](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/resource-schema.json) there.
Could we add a bit of context on which resource we are talking about?

detzgk:
LGTM, as @bcaudan pointed out, we could put this in the replay namespace or call it ReplayResourceSchema if we want to be more specific.

detzgk (dismissed):

BenoitZugmeyer:
Maybe we could rename "resource" to something else like "asset" to avoid confusion.

schemas/asset-schema.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "asset-schema.json",
+  "title": "AssetEvent",
```
BenoitZugmeyer: **💬 suggestion:** ‏ ‏is it an event? Maybe we could call this `AssetMetadata` like we do for segments.

schemas/asset-schema.json
```diff
@@ -0,0 +1,27 @@
...
+  "type": "object",
+  "description": "Schema of ",
+  "required": ["type", "application"],
+  "properties": {
+    "type": {
+      "const": "resource"
```
BenoitZugmeyer: **❓ question:** ‏Are we planning to have other types? If so, what would it look like? I'm asking, so we can be sure that the `asset` naming and this `type:resource` property make sense

schemas/asset-schema.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "asset-schema.json",
+  "title": "AssetEvent",
+  "type": "object",
+  "description": "Schema of ",
```
BenoitZugmeyer: **❓ question:** ‏Schema of... what?

detzgk (dismissed):
I'm still pretty iffy on calling this asset when it's resource everywhere else. I call them "replay resource" whenever I need to differentiate but usually I'm operating in some kind of replay namespace.

But I won't be a blocker about it.

maciejburda:
+1 @detzgk 

It's being called `Resource` by the product as well. Probably worth agreeing on the same nomenclature

schemas/asset-schema.json
ThibautGeriz: The other type is segment, as far as I know we do not plan to have other type

BenoitZugmeyer (dismissed):

BenoitZugmeyer (dismissed):

maciejburda (dismissed):
Thanks for considering a rename 🖖

schemas/resource-metadata-schema.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "resource-metadata-schema.json",
+  "title": "ResourceMetadata",
+  "type": "object",
+  "description": "Schema of the resource metadata. The metadata to idenfify the asset's content uploaded to th backend",
```
bcaudan:
```suggestion
  "description": "Schema of the resource metadata. The metadata to idenfify the asset's content uploaded to the backend",
```

schemas/resource-metadata-schema.json
bcaudan, line 1: **❓ question:** ‏would it worth putting it in the session-replay folder?

ThibautGeriz: In its form, it's not tight to replay (at least for the collection part). It could be used for some other purpose
That being said, this ain't a no strong opinion. I can move if you believe it's better

bcaudan: It is just a bit different than the other files that are at the root level.
I guess it should not be a big deal to move this file later if we feel the need.

maciejburda (approved):

bcaudan (approved):

PR #168, opened 2023-11-07
RUM-1837 Update telemetry schema

maciejburda:
Adds missing fields to configuration telemetry model

### DIFF
--- a/schemas/telemetry/configuration-schema.json
+++ b/schemas/telemetry/configuration-schema.json
@@ -288,6 +288,14 @@
                   "type": "integer",
                   "description": "The upload frequency of batches (in milliseconds)"
                 },
+                "batch_processing_level": {
+                  "type": "integer",
+                  "description": "Maximum number of batches processed sequencially without a delay"
+                },
+                "background_tasks_enabled": {
+                  "type": "boolean",
+                  "description": "Whether UIApplication background tasks are enabled"
+                },
                 "react_version": {
                   "type": "string",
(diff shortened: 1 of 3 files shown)


schemas/telemetry/configuration-schema.json
```diff
@@ -288,6 +288,14 @@
                   "type": "integer",
                   "description": "The upload frequency of batches (in milliseconds)"
                 },
+                "batch_processing_level": {
+                  "type": "integer",
+                  "description": "Maximum number of batches processed sequencially without a delay"
+                },
+                "background_tasks_enabled": {
+                  "type": "boolean",
+                  "description": "Whether UIApplication background tasks are enabled"
```
0xnm, line 297: this needs to be changed, it is not only about `UIApplication` :)

maciejburda: I think it's a configuration only available on iOS 🤔

mariusc83: yeah, I was about to ask what's that property all about ?

mariusc83 (dismissed):

schemas/telemetry/configuration-schema.json
maciejburda: This is something that we forgot to add the other day and I discovered it when working on `batch_processing_level`

I can move it to a separate PR too :)

mariusc83 (approved):

PR #170, opened 2023-11-14
RUM-2162: Add parent view id to event schema

N-Boutaib:
### Changes
Adding a new optional field that contains data about the parent view of RUM Events.
This new field is added on all RUM Events except SessionRumEvent.
This field has two attributes:

```typescript
parent_view: {
    id: string // uuid
    source: string // source of the parent view
}

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -8,7 +8,7 @@ export declare type RumEvent = RumActionEvent | RumErrorEvent | RumLongTaskEvent
 /**
  * Schema of all properties of an Action event
  */
-export declare type RumActionEvent = CommonProperties & {
+export declare type RumActionEvent = CommonProperties & ParentViewSchema & {
     /**
      * RUM event type
      */
@@ -150,7 +150,7 @@ export declare type RumActionEvent = CommonProperties & {
 /**
  * Schema of all properties of an Error event
  */
-export declare type RumErrorEvent = CommonProperties & ActionChildProperties & {
+export declare type RumErrorEvent = CommonProperties & ActionChildProperties & ParentViewSchema & {
     /**
      * RUM event type
      */
@@ -280,7 +280,7 @@ export declare type RumErrorEvent = CommonProperties & ActionChildProperties & {
 /**
  * Schema of all properties of a Long Task event
  */
-export declare type RumLongTaskEvent = CommonProperties & ActionChildProperties & {
+export declare type RumLongTaskEvent = CommonProperties & ActionChildProperties & ParentViewSchema & {
     /**
      * RUM event type
      */
@@ -318,7 +318,7 @@ export declare type RumLongTaskEvent = CommonProperties & ActionChildProperties 
 /**
  * Schema of all properties of a Resource event
  */
-export declare type RumResourceEvent = CommonProperties & ActionChildProperties & {
+export declare type RumResourceEvent = CommonProperties & ActionChildProperties & ParentViewSchema & {
     /**
      * RUM event type
      */
@@ -508,7 +508,7 @@ export declare type RumResourceEvent = CommonProperties & ActionChildProperties 
 /**
  * Schema of all properties of a View event
  */
-export declare type RumViewEvent = CommonProperties & {
+export declare type RumViewEvent = CommonProperties & ParentViewSchema & {
     /**
      * RUM event type
      */
@@ -1099,6 +1099,26 @@ export interface CommonProperties {
      * User provided context
      */
     context?: {
+        [k: string]: unknown;
+    };
+    [k: string]: unknown;
+}
+/**
+ * Parent view schema for views that are nested (webviews in mobile)
+ */
+export interface ParentViewSchema {
+    /**
+     * Parent view properties (view wrapping the current view)
+     */
+    readonly parent_view?: {
+        /**
+         * ID of the parent view
+         */
+        readonly id: string;
+        /**
+         * Source of the parent view
+         */
+        readonly source: 'android' | 'ios' | 'browser' | 'flutter' | 'react-native' | 'roku';
         [k: string]: unknown;
     };
     [k: string]: unknown;
--- a/schemas/rum/_parent-view-schema.json
+++ b/schemas/rum/_parent-view-schema.json
@@ -1,0 +1,29 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/_parent-view-schema.json",
+  "title": "ParentViewSchema",
+  "type": "object",
+  "description": "Parent view schema for views that are nested (webviews in mobile)",
+  "properties": {
+    "parent_view": {
+      "type": "object",
+      "description": "Parent view properties (view wrapping the current view)",
+      "required": ["id", "source"],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "ID of the parent view",
+          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+          "readOnly": true
+        },
+        "source": {
+          "type": "string",
+          "description": "Source of the parent view",
+          "enum": ["android", "ios", "browser", "flutter", "react-native", "roku"],
+          "readOnly": true
+        }
+      },
+      "readOnly": true
+    }
+  }
+}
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Review by mariusc83 — DISMISSED** [review]
LGTM

File: schemas/rum/view-schema.json [hunk 1]
```diff
@@ -323,6 +323,26 @@
           },
           "readOnly": true
         },
+        "parent_view": {
```
mquentin: forgot to do like other schema:
```
{
      "$ref": "_parent-view-schema.json"
    },
```

schemas/rum/_parent-view-schema.json
```diff
@@ -0,0 +1,29 @@
...
+        "id": {
+          "type": "string",
+          "description": "ID of the parent view",
+          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+          "readOnly": true
+        },
+        "source": {
+          "type": "string",
+          "description": "Source of the parent view",
+          "enum": ["android", "ios", "browser", "flutter", "react-native", "roku"],
```
mquentin, line 22: I see already too many 

`      "enum": ["android", "ios", "browser", "flutter", "react-native", "roku"],
`

Duplicated enums.

Suggestion:
create a shared enum

lib/cjs/generated/rum.d.ts
```diff
@@ -725,6 +725,20 @@ export declare type RumViewEvent = CommonProperties & {
         js_refresh_rate?: RumPerfMetric;
         [k: string]: unknown;
     };
+    /**
+     * Parent view properties (view wrapping the current view)
+     */
+    readonly parent_view?: {
```
BenoitZugmeyer: **💬 suggestion:** ‏On the replay side we have `nestedEnvId`. I think this is kind of the same concept, so maybe we could have a unified wording for both of those properties? like:

```
nested_env: {
   view: { id: string },
   source: ...
}
```

or `parent_env` (and we rename `nestedEnvId` to `parentEnvId` in replay)

mquentin: I don't have a strong opinion on this one but as we already have a parent hierarchy for RUM events (Session => view => error) maybe going all in for nested is better.

bcaudan: FMU, unlike `nestedEnvId` on the replay side, those new attributes on the RUM events would be exposed to customers.
I feel like the "nested env" concept may not be that clear to customers and that exposing a "parent view" concept could be easier to understand.

schemas/rum/view-schema.json
mariusc83: Or it could be added in the `_common-schema` ?

N-Boutaib: Done ✔️

lib/cjs/generated/rum.d.ts
N-Boutaib: I think it makes sense to have `parent_view` since we're adding it on the nested webview, while for replay `nested_env` is added on the wrapping wireframe (from the parent native view)

mquentin (approved):
LGTM with the enum change for later

lib/cjs/generated/rum.d.ts
BenoitZugmeyer: > those new attributes on the RUM events would be exposed to customers.

Good point. Maybe they don't need to?

> while for replay nested_env is added on the wrapping wireframe

No it's added to all Browser records too, not only the webview wireframe record: https://github.com/DataDog/rum-events-format/pull/165/files

bcaudan: > > those new attributes on the RUM events would be exposed to customers.

> Good point. Maybe they don't need to?

Let's say that you want to see all errors generated on a specific page / screen regardless of the source, it would allow to search for `@view.id: xxx OR @parent_view.id:xxx`.
Not sure that we need to support this kind of use case, it could bring some value though.

schemas/rum/_parent-view-schema.json
mariusc83: not sure our schema generator supports that now honestly. I would rather not add this new constraint to it.

mariusc83 (approved):

lib/cjs/generated/rum.d.ts
ThibautGeriz: loosly held opinion: I prefer to avoid `nestedEnv` and stick with `parent_view.id` like `initial_view.id`

BenoitZugmeyer: Proposal: use "parent" (as "nested" is not quite clear)

* In *Browser Records*, rename `nestedEnvId` to `parentEnvId`

* In *Browser RUM events*:
  * rename `parent_view.id` to `parent.view.id`
  * rename `parent_view.source` to `parent.source`
  * (maybe one day we'll have `parent.env.id` or something like that if needed)

N-Boutaib: I do agree with the first point (`nested` --> `parent`).
For the second point, I think `parent` is a bit abstract. Moreover, we add a second layer of nesting in `parent.view.id`. I would suggest we stick with the current naming strategy.
What do you think @bcaudan @mquentin @ThibautGeriz

mquentin: > I do agree with the first point (nested --> parent).

Same here I feel it's a better naming overall.

`nestedEnvId` => `parentEnvId` works great to me then. (just remember that it will break all the current Mobile Webview example we have and are currently debuging)

Like Thibaut said, we already started with the `initial_view` so `parent_view` sounds a good option to me. On top of that adding a later `parent_env` does not collide with `parent_view` as an env and a view will be different entities with different props

bcaudan: > In Browser Records, rename nestedEnvId to parentEnvId

Do we also have mentions of nestedEnvId on some mobile records?
If yes, maybe we could remove the "nested" or "parent" prefix from all the records.
FMU, the nested or parent aspect depends a bit on the point of view (from the parent or the child).

BenoitZugmeyer: > If yes, maybe we could remove the "nested" or "parent" prefix from all the records

This sounds good to me!

---

I see my proposal don't have much traction, so I'll write my thoughts and then feel free to merge. We are introducing a new customer facing concept here, and I think it's fine if we take a bit of time to find a good naming for it.

### Why I don't like having `view` in the newly introduced root namespace?

I proposed `parent.view.id` instead of `parent_view.id`, because we also want to add the parent `source` field, and the existing [`source` field](https://github.com/DataDog/rum-events-format/blob/d9624e702e86352babaf59979877f980eaccb8ff/lib/esm/generated/rum.d.ts#L891C5-L891C13) is a "root level" field (it's `@source`, not `@view.source`) so `@parent_view.source` does not make sense to me, and using a dedicated namespace to represent all the "parent environment" fields would be better. It would also allow to add more things unrelated to views into it, for example the "parent environment id".

I am aware of the `initial_view` property, and I agree that if we only had view-related fields, `parent_view` could be nice. But we don't.

### Maybe we need another name for that concept?

Maxime [noticed](https://github.com/DataDog/rum-events-format/pull/170#discussion_r1394524194) that the "parent" concept is already taken for the relationship between Session -> View -> Action -> Error / Resource / Longtask. It is customer facing as we are mentioning "Parent hierarchy" in the UI.

If we were introducing those new fields only on View events, it could have been fine, it could be seen as a new layer of the "parent hierarchy".

But those fields will be on all events. So a customer might wonder which field to use to target the parent view of an Error event: should they use `@view.id` or `@parent_view.id`?

Now, I understand `nested_env` is not quite clear. I proposed `parent_env`, but maybe we can think of something else?

bcaudan: > We are introducing a new customer facing concept here, and I think it's fine if we take a bit of time to find a good naming for it.

👍 super interesting discussion

> I proposed parent.view.id instead of parent_view.id, because we also want to add the parent source field

👍 That makes sense to me too

> the "parent" concept is already taken for the relationship between Session -> View -> Action -> Error / Resource / Longtask.

What about something like "container" or "host" instead of parent then?

Other thought: 
if we plan to expose the "env" concept on the RUM side, it would also collides with the existing one (service/env/version) 😕
Renaming it to something like "runtime" could work.

So it could look like
- mobile webview record: runtimeId to identify the id of the corresponding webview
- browser record: runtimeId to identify the id of the webview they are in
- browser rum events: container_runtime.* to attach data from the mobile side (view, source)

PR #176, opened 2024-01-03
Validate objects properties case

BenoitZugmeyer:
Most objects defined by our schemas have consistent casing, but it turns out there are some [unexpected](https://github.com/DataDog/rum-events-format/blob/e792ad1d75bd4a375729f37764cd44a3f9233fb0/schemas/session-replay/common/focus-record-schema.json#L24) [exceptions](https://github.com/DataDog/rum-events-format/blob/e792ad1d75bd4a375729f37764cd44a3f9233fb0/schemas/rum/resource-schema.json#L235-L245).

To try to keep casing consistent in the future, this PR changes the validation script to make sure all properties are cased correctly.


scripts/validate.mjs
```diff
@@ -2,40 +2,76 @@ import path from 'path'
...
+    const schema = readJson(schemaPath)
 
-  ajv.addSchema(schema)
+    // RUM and telemetry schemas object properties should be snake_case, other schemas objects should
+    // be camelCase
+    const shouldBeSnakeCase =
+      schemaPath.startsWith(`${SCHEMAS_DIRECTORY}/rum/`) || schemaPath.startsWith(`${SCHEMAS_DIRECTORY}/telemetry/`)
+
+    // Some exceptions to the rule above. Ideally they should be fixed in the future.
+    const caseExceptions =
```
anediaz: suggestion:
I'd avoid embedded ternaries for readability concerns. Instead, I'd rather use a constant where all exceptions are gathered together, something like this:

```
const CASE_EXCEPTIONS = {
  [`${SCHEMAS_DIRECTORY}/session-replay/common/_common-segment-metadata-schema.json`]: ['records_count', 'index_in_view', 'has_full_snapshot'],
  [`${SCHEMAS_DIRECTORY}/session-replay/common/focus-record-schema.json`]: ['has_focus'],
  [`${SCHEMAS_DIRECTORY}/rum/resource-schema.json`]: ['operationType', 'operationName']
}
```

and then assign the correct value to your constant:
`const caseExceptions = CASE_EXCEPTIONS[schemaPath] || [];`

scripts/validate.mjs
```diff
@@ -64,3 +100,34 @@ function forEachFile(directoryPath, callback) {
 function readJson(filePath) {
   return JSON.parse(readFileSync(filePath, 'utf8'))
 }
+
+/**
+ * Iterates over each properties of objects specified in the provided JSON schema.
+ */
+function forEachObjectProperty(schema, callback) {
+  if (Array.isArray(schema)) {
```
anediaz, line 119: **suggestion:**
you could early return here if `schema` is `null`, this will simplify a little bit the rest of the code

scripts/validate.mjs
```diff
@@ -64,3 +100,34 @@ function forEachFile(directoryPath, callback) {
...
+    for (const value of schema) {
+      forEachObjectProperty(value, callback)
+    }
+  } else if (typeof schema === 'object' && schema !== null) {
+    // traverse objects
+    for (const value of Object.values(schema)) {
+      forEachObjectProperty(value, callback)
+    }
+
+    if (schema.type === 'object' && schema.properties) {
```
anediaz, line 130: **question:**
when `properties` exists in `schema`, it can only be an object? shouldn't you check if `properties` is different of `null` in case of?

scripts/validate.mjs
```diff
@@ -2,40 +2,76 @@ import path from 'path'
 import Ajv from 'ajv'
```
anediaz, line 2: **chore:**
not sure if it exists in `.mjs` files but, it'd be nice to have some unit tests 😇

scripts/validate.mjs
```diff
@@ -2,40 +2,76 @@ import path from 'path'
...
-    console.log(`⚠️ Schema ${schemaPath} $id should be ${schemaId}`)
-    process.exitCode = 1
-  }
+function validateSchemasObjectsPropertiesCase() {
+  forEachFile(SCHEMAS_DIRECTORY, (schemaPath) => {
+    const schema = readJson(schemaPath)
 
-  ajv.addSchema(schema)
+    // RUM and telemetry schemas object properties should be snake_case, other schemas objects should
+    // be camelCase
```
louiszawadzki, line 33: Question: would it be possible to have consistent casing across all schemas?

scripts/validate.mjs
```diff
@@ -2,40 +2,76 @@ import path from 'path'
...
+        : schemaPath === `${SCHEMAS_DIRECTORY}/session-replay/common/focus-record-schema.json`
+        ? ['has_focus']
+        : schemaPath === `${SCHEMAS_DIRECTORY}/rum/resource-schema.json`
+        ? ['operationType', 'operationName']
+        : []
+
+    forEachObjectProperty(schema, (key) => {
+      const isCorrectCase = shouldBeSnakeCase ? isSnakeCase(key) : isCamelCase(key)
+      if (!isCorrectCase && !caseExceptions.includes(key)) {
+        console.log(`❌ Schema ${schemaPath} property ${key} is not ${shouldBeSnakeCase ? 'snake_case' : 'camelCase'}`)
```
louiszawadzki, line 42: Add clarification if you've looked into another schemas to find the correct casing.

```suggestion
        console.log(`❌ Schema ${schemaPath} property ${key} is not ${shouldBeSnakeCase ? 'snake_case' : 'camelCase'}`)
        console.log("RUM and Telemetry schemas should be snake_case, other schemas should be camelCase.")
```

louiszawadzki (dismissed):
I've added some questions/suggestions to clarify the expected casing depending on the schema location.
It's minor, can be merged without it :)

scripts/validate.mjs
BenoitZugmeyer: Yes `properties` is always an object: https://json-schema.org/understanding-json-schema/reference/object#properties

scripts/validate.mjs
BenoitZugmeyer: I don't see how it would simplify the code?

The `schema !== null` is directly related to `typeof schema === 'object'` (since `typeof null === 'object'`). By hoisting this condition, you would need to remember that `schema` cannot be `null` when coming across the `typeof schema === 'object'` because of the early exit.

I think keeping those two conditions together makes sense here.

anediaz: it is really a nitpick, nothing blocking. I get what you're doing in this part of the code, I just thought having something like this could simplify if (basically to simplify the if/else):

```
if (schema === null){
  return;
}
if (Array.isArray(schema)) {
  ....
}
else if (typeof schema === 'object') {

}
```

scripts/validate.mjs
BenoitZugmeyer: It would be a very significant task. Changing snake_case to camelCase in RUM/Telemetry events would break pretty much everything including customer queries. Changing camelCase to snake_case in session replay is more easily achievable as it is not customer facing and to my knowledge only used in the frontend, but it would require @DataDog/rum-session-replay to handle both case styles (earlier versions of SDKs would still use camelCase) pretty much everywhere across the player and associated toolings.

scripts/validate.mjs
BenoitZugmeyer: Good suggestion. I added integration tests (i.e. running the command and checking its output) instead of unit tests as it will be simpler to cover bigger scope than unit tests. Let me know what you think!

anediaz: cool! 🚀

anediaz (approved):

amortemousque (approved):

PR #177, opened 2024-01-09
[RUM-2729] Update connectivity model with browser support

bcaudan:
# Motivation

Collect connectivity data for RUM browser

# Changes

- Make `connectivity.interfaces` optional: not supported by every browser
- Add `connectivity.effective_type` (`slow-2g`, `2g`, `3g`, `4g`)

### DIFF
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -132,7 +132,7 @@
     "connectivity": {
       "type": "object",
       "description": "Device connectivity properties",
-      "required": ["status", "interfaces"],
+      "required": ["status"],
       "properties": {
         "status": {
           "type": "string",
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -149,6 +149,12 @@
           },
           "readOnly": true
         },
+        "effective_type": {
+          "type": "string",
+          "description": "Effective type of the network connection",
+          "enum": ["slow_2g", "2g", "3g", "4g"],
+          "readOnly": true
+        },
         "cellular": {
           "type": "object",
           "description": "Cellular connectivity properties",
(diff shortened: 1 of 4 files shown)


amortemousque (dismissed):

schemas/rum/_common-schema.json
```diff
@@ -149,6 +149,12 @@
           },
           "readOnly": true
         },
+        "effective_type": {
+          "type": "string",
+          "description": "Effective type of the network connection",
+          "enum": ["slow_2g", "2g", "3g", "4g"],
+          "readOnly": true
+        },
```
xgouchet: On Mobile, we're using the `cellular.technology` attribute to display the type of mobile network connection. What's the purpose of the additional "effective_type" ? Also don't you also need non mobile connection types (ADSL, Fiber, …)?

bcaudan: Looking at what is collected in `cellular.technology`, I don't think we have an API to retrieve something similar.
Effective type gives an indication of the speed of the network, the listed values are the ones provided by the browser API.
cf [NetworkInformation](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation) for more details on what is available on the browser side.

xgouchet: Ok, the naming is a bit ambiguous, maybe the documentation could reflect better the meaning, e.g.: 
> Bucket describing the speed and bandwidth of the network connection based on measurements of mobile data connections

bcaudan: Updated with:
> Cellular connection type reflecting the measured network performance

to stay close to the [mdn doc wording](https://developer.mozilla.org/en-US/docs/Glossary/Effective_connection_type)

BenoitZugmeyer (approved):

xgouchet (approved):

PR #182, opened 2024-02-01
[RUM-2889] Add vital schema

bcaudan:
Specify new vital event schema for the custom vital feature:

```js
{
  "date": number,
  "type": "vital",
  "vital" {
    "id": string,
    "type": "duration"
    "custom": {
      "my-timing": number,
    }
  }
}
```

### DIFF
--- a/samples/rum-events/vital.json
+++ b/samples/rum-events/vital.json
@@ -1,0 +1,27 @@
+{
+  "type": "vital",
+  "date": 1706792746576,
+  "application": {
+    "id": "ac8218cf-498b-4d33-bd44-151095959547"
+  },
+  "session": {
+    "id": "cacbf45c-3a05-48ce-b066-d76349460599",
+    "type": "user"
+  },
+  "source": "browser",
+  "view": {
+    "id": "623d50fd-75cf-4025-97d2-e51ff94171f6",
+    "referrer": "",
+    "url": "https://app.datadoghq.com/rum/explorer?live=1h&query=&tab=view"
+  },
+  "vital": {
+    "id": "47af0abb-3de1-4170-a491-4473716df979",
+    "type": "duration",
+    "custom": {
+      "user-timing": 100
+    }
+  },
+  "_dd": {
+    "format_version": 2
+  }
+}
--- a/schemas/rum/vital-schema.json
+++ b/schemas/rum/vital-schema.json
@@ -1,0 +1,56 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/vital-schema.json",
+  "title": "RumVitalEvent",
+  "type": "object",
+  "description": "Schema of all properties of a Vital event",
+  "allOf": [
+    {
+      "$ref": "_common-schema.json"
+    },
+    {
+      "$ref": "_view-container-schema.json"
+    },
+    {
+      "required": ["type", "vital"],
+      "properties": {
+        "type": {
+          "type": "string",
+          "description": "RUM event type",
+          "const": "vital",
+          "readOnly": true
+        },
+        "vital": {
+          "type": "object",
+          "description": "Vital properties",
+          "required": ["id", "type"],
+          "properties": {
+            "type": {
+              "type": "string",
+              "description": "Type of the vital",
+              "enum": ["duration"],
+              "readOnly": true
+            },
+            "id": {
+              "type": "string",
+              "description": "UUID of the vital",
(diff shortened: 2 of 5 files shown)


samples/rum-events/vital.json
```diff
@@ -0,0 +1,27 @@
+{
+  "type": "vital",
+  "date": 1706792746576,
```
anediaz, line 3: **question:**
this corresponds to the date where the custom vital has been created?

schemas/rum/vital-schema.json
```diff
@@ -0,0 +1,56 @@
...
+            },
+            "id": {
+              "type": "string",
+              "description": "UUID of the vital",
+              "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+              "readOnly": true
+            },
+            "custom": {
+              "type": "object",
+              "description": "User custom vital. As vital name is used as facet path, it must contain only letters, digits, or the characters - _ . @ $",
```
BenoitZugmeyer, line 42: **❓ question:** ‏ ‏if it contains `.`, the field will be nested in an object (as `custom['foo.bar'] = 42` is equivalent to `custom.foo = { bar: 42 }` once processed). Is that intented?

If it is, we probably should adapt the type to something like:
```ts
custom: CustomVitalValue

type CustomVitalValue = number | Record<string, CustomVitalValue>
```
so we can either set it with a `.` or a nested object (like other Context)

anediaz:
> Specify new vital event schema for the custom vital feature:
> 
> ```js
> {
>   "date": number,
>   "type": "vital",
>   "vital" {
>     "id": string,
>     "type": "duration"
>     "custom": {
>       "my-timing": number,
>     }
>   }
> }
> ```

**question:**
I wonder if we could add in the vital definition its unit name (`seconds`, `nanoseconds`...). This is something important to know when manipulating values in the frontend.
Is it maybe expected in the `custom` field?

xgouchet (approved):

samples/rum-events/vital.json
bcaudan: It corresponds to the start of the vital.

anediaz: oh my bad :-( of course!

schemas/rum/vital-schema.json
bcaudan: I copy-pasted the description from `view.custom_timings` without thinking too much about it.

From a SDK / ingestion perspective, the type should only be `number`.
From the app perspective, it will indeed be retrieved an an object but from a quick test, a `foo.bar` custom timing seems to be handled properly.

Since the types documented here are the ones sent by the SDK, I would tend to leave it as it is.
For the app team, it worth noting that the `.` should be handled like it is for `custom_timings`, cc @anediaz. 

wdyt?

samples/rum-events/vital.json
bcaudan: No worries ;)

bcaudan:
> **question:** I wonder if we could add in the vital definition its unit name (`seconds`, `nanoseconds`...). This is something important to know when manipulating values in the frontend. Is it maybe expected in the `custom` field?

Like for other customers fields, the idea is to delegate the unit configuration to the facet rather than creating a new unit system inside the event attributes.

anediaz (approved):

schemas/rum/vital-schema.json
BenoitZugmeyer: Ok. To me, it doesn't make sense to allow one way but not the other to send nested properties. I'm not sure we need to be similar to custom timings in this case. But it might be fine as long as other involved teams are aware of this.

BenoitZugmeyer (approved):

schemas/rum/vital-schema.json
bcaudan: Discussed IRL, allowing object there does not seem to address any use case and it was surely unintentional for custom timings.
Preventing the use of `.` could avoid unnecessary logic to handle this case either on app or backend side.

wdyt @anediaz and @kbogtob?

anediaz: if I understand correctly, Browser SDK will prevent sending nested properties, and this will simplify treatments in the frontend and the backend right?

bcaudan: yes, the idea would to be to not let customers use `.` for vital name in order to avoid the handle the object conversion done by the backend.

kbogtob (approved):

schemas/rum/vital-schema.json
bcaudan: I'll go ahead and merge this PR to not block depending tasks and move this discussion to a RFC.

PR #186, opened 2024-02-27
IW: metric event format

bcaudan:
**proposal:**
```
{
  "type": "metric",
  "metric": {
    "type": "series",
    "series": [
      ...
    ]
  }
}
```
to allow to have a [distribution_points](https://docs.datadoghq.com/api/latest/metrics/#submit-distribution-points) type later if we want.


maxep:
Nice! I would place it in the [schemas](https://github.com/DataDog/rum-events-format/tree/master/schemas) folder as it's not related to rum, WDYT?

PR #188, opened 2024-03-06
RUM-2925 Add error properties for iOS App Hangs and Android ANRs monitoring

ncreated:
### What and Why?

📦 This PR updates error and configuration schemas for iOS App Hang and Android ANR errors monitoring.

### How?

Added following properties:

**`error.category`** - This is to categorize mobile errors and offer a facet for high-level grouping with certain categories. We start with 3 categories: `App Hang` | `ANR` | `Exception`, but more will come later. For more context see [this RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3475931379/RFC+-+ANR+and+App+Hangs+in+Mobile+RUM#Implement-distinct-error-category-for-hangs%3A-error.category-facet) (internal).

**`error.freeze.duration`** - To send the duration of the main thread freeze in ANR or App Hang. More context in [this fragment of the RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3475931379/RFC+-+ANR+and+App+Hangs+in+Mobile+RUM#Track-duration-of-non-fatal-hangs) (internal).

**`telemetry.app_hang_threshold`** - To measure the usage of App Hangs monitoring in iOS.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -210,6 +210,10 @@ export declare type RumErrorEvent = CommonProperties & ActionChildProperties & V
          */
         readonly type?: string;
         /**
+         * The specific category of the error. It provides a high-level grouping for different types of errors.
+         */
+        readonly category?: 'ANR' | 'App Hang';
+        /**
          * Whether the error has been handled manually in the source code or not
          */
         readonly handling?: 'handled' | 'unhandled';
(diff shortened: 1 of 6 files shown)


0xnm (dismissed):

0xnm (approved):

lib/cjs/generated/rum.d.ts
```diff
@@ -349,6 +353,16 @@ export declare type RumErrorEvent = CommonProperties & ActionChildProperties & V
...
     };
+    /**
+     * Properties of App Hang and ANR errors
+     */
+    readonly freeze?: {
+        /**
+         * Duration of the main thread freeze (in ns)
+         */
+        readonly duration: number;
+        [k: string]: unknown;
```
maciejburda, line 364: just curious - why is that `[k: string]: unknown`  needed?

maciejburda (approved):

lib/cjs/generated/rum.d.ts
ncreated: AFAIK it means the object can have any number of additional properties with keys of type `string` and values of any type (`unknown`). Looks that it generates "additional properties" by default (if not opted-out in the schema definition). This object is not meant to be used in Browser SDK, hence it should be no problem.

PR #193, opened 2024-03-20
RUM-3531 Add `isVisible` to webview wireframe

maxep:
## Session Replay WebView

Report if a webview wireframe in memory is visible or not.

### DIFF
--- a/schemas/session-replay/mobile/webview-wireframe-schema.json
+++ b/schemas/session-replay/mobile/webview-wireframe-schema.json
@@ -21,6 +21,11 @@
           "type": "string",
           "description": "Unique Id of the slot containing this webview.",
           "readOnly": true
+        },
+        "isVisible": {
+          "type": "boolean",
+          "description": "Whether this web-view is visible or not.",
+          "readOnly": true
         }
       }
     }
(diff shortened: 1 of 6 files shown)


schemas/session-replay/mobile/webview-wireframe-schema.json
```diff
@@ -21,6 +21,11 @@
           "type": "string",
           "description": "Unique Id of the slot containing this webview.",
           "readOnly": true
+        },
+        "hasFocus": {
```
xgouchet: > [!IMPORTANT]
> It feels that the notion of focus is different on mobile than it is on the Web. You could have two webviews on the same screen, both visible, but only one which has focus. I would advise to use `isVisible` instead, as it's more accurate to what happens on the mobile.

xgouchet (changes requested):

schemas/session-replay/mobile/webview-wireframe-schema.json
maxep: Changed 👍

mquentin (approved):
LGTM

xgouchet (approved):
Nice, thanks

PR #200, opened 2024-04-18
RUM-1844: Add device and os info to telemetry events

jonathanmos:
Adds device and OS information to telemetry events. While we have some device and os information inside http_agent, we require this set of properties for performance metrics

### DIFF
--- a/schemas/telemetry/_common-schema.json
+++ b/schemas/telemetry/_common-schema.json
@@ -100,6 +100,48 @@
         "type": "string"
       },
       "readOnly": true
+    },
+    "telemetry": {
+      "type": "object",
+      "additionalProperties": true,
+      "properties": {
+        "device": {
+          "type": "object",
+          "description": "Device properties",
+          "properties": {
+            "architecture": {
+              "type": "string",
+              "description": "Architecture of the device"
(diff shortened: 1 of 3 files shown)


xgouchet (approved):

schemas/telemetry/_common-schema.json
```diff
@@ -100,6 +100,48 @@
         "type": "string"
       },
       "readOnly": true
+    },
+    "telemetry": {
+      "type": "object",
+      "additionalProperties": true,
```
ncreated, line 106: question/ Why also adding `"additionalProperties": true`? It looks unrelated to `device` and `os` change.

jonathanmos: because the error-schema and debug-schema also add properties to the telemetry object, so this is to enable a merge

ncreated: Got it 👍. Checking if this is not added by accident. Makes sense 👍.

ncreated (approved):

RomanWuattier (approved):

bcaudan (approved):
LGTM

PR #202, opened 2024-04-26
[RUM] Add force session replay option to telemetry usage features

N-Boutaib:
## Motivation
Since we are adding a new way to force the recording of sampled out sessions, it would be useful to keep track of the usage of this API.

## Changes
Adding a new `force-session-replay-recording` option to report when the recorder is forced to start on a sampled out session.

### DIFF
--- a/schemas/telemetry/usage/browser-features-schema.json
+++ b/schemas/telemetry/usage/browser-features-schema.json
@@ -6,7 +6,7 @@
   "description": "Schema of browser specific features usage",
   "oneOf": [
     {
-      "required": ["feature", "is_forced"],
+      "required": ["feature"],
       "properties": {
         "feature": {
           "type": "string",
--- a/schemas/telemetry/usage/browser-features-schema.json
+++ b/schemas/telemetry/usage/browser-features-schema.json
@@ -6,12 +6,17 @@
   "description": "Schema of browser specific features usage",
   "oneOf": [
     {
-      "required": ["feature"],
+      "required": ["feature", "start_type"],
       "properties": {
         "feature": {
           "type": "string",
           "description": "startSessionReplayRecording API",
           "const": "start-session-replay-recording"
+        },
+        "start_type": {
+          "type": "string",
+          "description": "Whether the recording was started normally or forced",
+          "enum": ["normal", "forced"]
         }
       }
     }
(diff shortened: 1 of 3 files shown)


schemas/telemetry/usage/browser-features-schema.json
```diff
@@ -14,6 +14,16 @@
...
         }
       }
+    },
+    {
+      "required": ["feature"],
+      "properties": {
+        "feature": {
+          "type": "string",
+          "description": "startSessionReplayRecording API",
+          "const": "force-session-replay-recording"
```
bcaudan: **💬 suggestion:** ‏since we are already monitoring the start API, what about adding the new parameter to it instead of adding a new feature?
(cf consent tracking to see how it is done)

N-Boutaib: ✔️ Adressed

xgouchet (dismissed):

schemas/telemetry/usage/browser-features-schema.json
```diff
@@ -6,12 +6,16 @@
...
+      "required": ["feature", "is_forced"],
       "properties": {
         "feature": {
           "type": "string",
           "description": "startSessionReplayRecording API",
           "const": "start-session-replay-recording"
+        },
+        "is_forced": {
+          "type": "boolean",
+          "description": "Whether the API was forced to record sampled out sessions"
```
bcaudan: **💬 suggestion:** the phrasing seems a bit confusing to me, wdyt of
‏
```suggestion
          "description": "Whether the recording is allowed to start even on sessions sampled out of replay"
```

xgouchet (dismissed):

schemas/telemetry/usage/browser-features-schema.json
```diff
@@ -6,12 +6,16 @@
   "description": "Schema of browser specific features usage",
   "oneOf": [
     {
-      "required": ["feature"],
+      "required": ["feature", "is_forced"],
```
hugolevyy: If this field is `required`, is it possible to have retro-compatibility issues ?

hugolevyy (dismissed):

schemas/telemetry/usage/browser-features-schema.json
bcaudan: Indeed, we could also avoid to send this parameter when the value is false

bcaudan (approved):

hugolevyy (approved):

PR #210, opened 2024-06-06
[RUM-4781] Add details and duration to vitals

nulrich:
- Duration to store the vital duration
- Details to store a secondary identifier for the vital

### DIFF
--- a/schemas/rum/vital-schema.json
+++ b/schemas/rum/vital-schema.json
@@ -42,6 +42,16 @@
               "description": "Name of the vital, as it is also used as facet path for its value, it must contain only letters, digits, or the characters - _ . @ $",
               "readOnly": true
             },
+            "details": {
+              "type": "string",
+              "description": "Details of the vital",
+              "readOnly": true
+            },
+            "duration": {
+              "type": "number",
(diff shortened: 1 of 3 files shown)


schemas/rum/vital-schema.json
```diff
@@ -42,6 +42,16 @@
               "description": "Name of the vital, as it is also used as facet path for its value, it must contain only letters, digits, or the characters - _ . @ $",
               "readOnly": true
             },
+            "details": {
+              "type": "string",
+              "description": "Details of the vital",
```
N-Boutaib: **Question**: by details you mean a description of what the vital is about ?

BenoitZugmeyer (dismissed):

schemas/rum/vital-schema.json
nulrich: The field is used to store for example a secondary identifier. Which will be used to track vitals for React component.

N-Boutaib: **Thought**: Could you please reflect that in the description ? This one is a bit ambiguous 😅

N-Boutaib (approved):

BenoitZugmeyer (approved):

PR #211, opened 2024-06-13
🏷️ [RUM-4469] add plugins to the configuration telemetry schema

BenoitZugmeyer:
In the Browser SDK we want to experiment with a plugin concept. Add it to the configuration telemetry so we can be aware of plugins adoption.


N-Boutaib (approved):
Nice ! Looking forward to adding plugins in the sdk

amortemousque (approved):

PR #229, opened 2024-11-13
RUM-7110 Add the OOB mobile view loading metrics properties

mariusc83:
Following our internal RFC we are adding 2 new properties on the `view-schema.json` to be used for the out-of-the-box view loading time metrics we propose for the our mobile RUM sdks.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -734,6 +734,14 @@ export declare type RumViewEvent = CommonProperties & ViewContainerSchema & {
          */
         readonly loading_time?: number;
         /**
+         * Duration in ns from the moment the view was started until all the initial network requests settled
+         */
+        readonly network_settled_time?: number;
+        /**
+         * Duration in ns to from the last interaction on previous view to the moment the current view was displayed
+         */
+        readonly interaction_to_next_view_time?: number;
+        /**
          * Type of the loading of the view
          */
         readonly loading_type?: 'initial_load' | 'route_change' | 'activity_display' | 'activity_redisplay' | 'fragment_display' | 'fragment_redisplay' | 'view_controller_display' | 'view_controller_redisplay';
--- a/schemas/rum/view-schema.json
+++ b/schemas/rum/view-schema.json
@@ -31,6 +31,18 @@
               "minimum": 0,
               "readOnly": true
             },
+            "network_settled_time": {
+              "type": "integer",
+              "description": "Duration in ns from the moment the view was started until all the initial network requests settled",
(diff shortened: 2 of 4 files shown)


schemas/rum/view-schema.json
```diff
@@ -31,6 +31,18 @@
               "minimum": 0,
               "readOnly": true
             },
+            "network_settled_time": {
+              "type": "integer",
+              "description": "Duration in ns from the moment the initial network requests in current view started until they settled",
```
xgouchet: > [!NOTE]
> IIRC from the RFC, the metric is measured from the moment the current view started, right, not from the start of the initial request?

```suggestion
              "description": "Duration in ns from the moment the current view started until the initial network requests settled",
```

schemas/rum/view-schema.json
```diff
@@ -31,6 +31,18 @@
...
+              "minimum": 0,
+              "readOnly": true
+            },
+            "interaction_to_next_view_time": {
+              "type": "integer",
+              "description": "Duration in ns to from the last interaction on previous view to the moment the next view is displayed",
```
xgouchet: > [!NOTE]
> The metric relates to the current view, not the next (future) one. 

```suggestion
              "description": "Duration in ns to from the last interaction on previous view to the moment the current view was displayed",
```

ncreated (dismissed):

schemas/rum/view-schema.json
mariusc83: yes correct, my mistake here 👍

schemas/rum/view-schema.json
mariusc83: yeah...for me `next` is actually `current` in the context. But if you think `current` is better I can change it yes

xgouchet (approved):

joaooliveira-datadog (approved):

lib/cjs/generated/rum.d.ts
```diff
@@ -733,6 +733,14 @@ export declare type RumViewEvent = CommonProperties & ViewContainerSchema & {
          * Duration in ns to the view is considered loaded
          */
         readonly loading_time?: number;
+        /**
+         * Duration in ns from the moment the view was started until all the initial network requests settled
+         */
+        readonly network_settled_time?: number;
+        /**
+         * Duration in ns to from the last interaction on previous view to the moment the current view was displayed
```
BenoitZugmeyer, line 741:
```suggestion
         * Duration in ns from the last interaction on previous view to the moment the current view was displayed
```

BenoitZugmeyer (approved):

PR #231, opened 2024-11-18
✨ [RUM-6956] Add name source to @_dd.action

cy-moi:
Add `name_source` field to `@_dd.action` related to PR https://github.com/DataDog/browser-sdk/pull/3115

### DIFF
--- a/schemas/rum/action-schema.json
+++ b/schemas/rum/action-schema.json
@@ -194,7 +194,7 @@
                 },
                 "name_source": {
                   "type": "string",
-                  "description": "Action name computing source",
+                  "description": "The strategy of how the auto click action name is computed",
                   "enum": [
                     "custom_attribute",
                     "mask_placeholder",
--- a/schemas/rum/action-schema.json
+++ b/schemas/rum/action-schema.json
@@ -191,6 +191,18 @@
                       "readOnly": true
                     }
                   }
+                },
+                "name_source": {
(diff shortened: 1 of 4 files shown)


schemas/rum/action-schema.json
```diff
@@ -191,6 +191,19 @@
                       "readOnly": true
                     }
                   }
+                },
+                "name_source": {
+                  "type": "string",
+                  "description": "Action name computing source",
```
xgouchet: > [!NOTE]
> Can we make the description a bit clearer as to what this attribute is for ? 
> Also was there any investigation of what enum values might be needed in the Mobile SDKs ?

cy-moi: Yes. I will add a clearer description. The enum types are discussed in this [RFC](https://docs.google.com/document/d/1JMhuwp3As7KJLC3NwWKyiB7R655d8yPXLAdhr81kwq8/edit?tab=t.0). This feature is experimental. We could also leave it a non-enum string as it is `readOnly: false`. Wdyt?

xgouchet: I think it maks sense to have an enum, I'll ask Mobile engineers working on SR privacy to take a look and add any missing value we might need.

BenoitZugmeyer (approved):

xgouchet (approved):

PR #240, opened 2024-12-23
🎨 [RUM-6567] Add new fields for Web Vitals attribution and other performance data

sethfowler-datadog:
This PR adds a new `RumPerformanceData` subobject within `RumViewEvent`. We'll gradually migrate performance-related fields from `RumViewEvent` into this subobject, taking the opportunity to reorganize them in a cleaner, more hierarchical fashion.

We begin the migration process within this PR by creating new versions of the Web-Vitals-related fields on `RumViewEvent` that live in `RumViewEvent`. In the new structure, these fields are organized hierarchically, so that all of the fields for each Web Vital live on a subobject and attribution data for each Web Vital is cleanly tied to the Web Vital's primary metric.

PR #245, opened 2025-01-13
🐛 [RUM-6567] Move performance namespace inside view

sethfowler-datadog:
This PR moves the `performance` namespace added in #240 inside `@view`, to align better with the pattern we've followed in the past for fields that are only used in the context of a single event.

Since we haven't yet landed the PRs that use the `performance` namespace in the `browser-sdk` repo, it's still possible to make this change safely.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -973,6 +973,10 @@ export declare type RumViewEvent = CommonProperties & ViewContainerSchema & {
          * The JavaScript refresh rate for React Native
          */
         js_refresh_rate?: RumPerfMetric;
+        /**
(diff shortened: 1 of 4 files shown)


BenoitZugmeyer (dismissed):

BenoitZugmeyer (commented):
You should also update the sample [here](https://github.com/DataDog/rum-events-format/blob/61e86d10ac47befa6691662778a25a43a8b1747a/samples/rum-events/view.json#L103)

thomas-lebeau (approved):

xgouchet (approved):

BenoitZugmeyer (approved):

PR #248, opened 2025-01-31
Fix effective_type format

amortemousque:
Fix effective_type format: `slow_2g` => `slow-2g`


BenoitZugmeyer (approved):

hemeryar:
Checked the backend already parses slow-2g, which I imagine is the reason for the fix 👍

hemeryar (approved):

PR #270, opened 2025-05-13
RUM-9561 Update view schema to add accessibility attributes

barboraplasovska:
Adds new accessibility parameters to the RUM View schema.

- `text_size` - User’s preferred text scale relative to the default system size.
- `screen_reader_enabled` - Indicates whether a screen reader is currently active.
- `bold_text_enabled` - Indicates whether the system-wide bold text accessibility setting is enabled.
- `reduce_transparency_enabled` - Indicates whether the system-wide reduce transparency setting is enabled.
- `reduce_motion_enabled` - Indicates whether the system-wide reduce motion setting is enabled.
- `button_shapes_enabled` - Indicates whether the system-wide button shapes setting is enabled.
- `invert_colors_enabled` - Indicates whether the system-wide color inversion setting is enabled.
- `increase_contrast_enabled` - Indicates whether the system-wide increase contrast setting is enabled.
- `assistive_switch_enabled` - Indicates whether an alternative input method like Switch Control or Switch Access is currently enabled.
- `assistive_touch_enabled` - Indicates whether the system-wide AssistiveTouch feature is currently enabled.
- `is_video_autoplay_enabled` - Indicates whether the video autoplay setting is enabled in the system or application.
- `closed_captioning_enabled` - Indicates whether closed captioning is enabled for media playback.
- `mono_audio_enabled` - Indicates whether the system-wide mono audio setting is enabled.
- `shake_to_undo_enabled` - Indicates whether the Shake to Undo feature is enabled.
- `reduced_animations_enabled` - Indicates whether the user prefers reduced animations or cross-fade transitions.
- `should_differentiate_without_color` - Indicates whether the system should differentiate interface elements without relying solely on color.
- `grayscale_enabled` - Indicates whether the device display is currently using grayscale mode.
- `single_app_mode_enabled` - Indicates whether the device is currently locked to a single app through Guided Access or Screen Pinning.
- `on_off_switch_labels_enabled` - Indicates whether on/off switch labels are enabled in the system settings.
- `speak_screen_enabled` - Indicates whether the Speak Screen feature is enabled.
- `speak_selection_enabled` - Indicates whether the text-to-speech selection feature is enabled.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -1007,6 +1007,10 @@ export declare type RumViewEvent = CommonProperties & ViewContainerSchema & {
          * Performance data. (Web Vitals, etc.)
          */
         performance?: ViewPerformanceData;
+        /**
+         * Accessibility properties of the view
+         */
+        accessibility?: ViewAccessibilityProperties;
         [k: string]: unknown;
     };
     /**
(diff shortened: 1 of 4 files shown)


hemeryar (dismissed):
lgtm

Manu-Tran (dismissed):

lib/cjs/generated/rum.d.ts
```diff
@@ -1740,3 +1744,93 @@ export interface RumRect {
...
+     */
+    readonly bold_text_enabled?: boolean;
+    /**
+     * Indicates whether the system-wide reduce transparency setting is enabled.
+     */
+    readonly reduce_transparency_enabled?: boolean;
+    /**
+     * Indicates whether the system-wide reduce motion setting is enabled.
+     */
+    readonly reduce_motion_enabled?: boolean;
```
mormubis, line 1770: **❓ question:** ‏ What is the difference between `reduce_motion_enabled` and `reduced_animations_enabled`? In the browser, we expect animations to be reduced or removed when we detect the `prefers-reduced-motion` flag.

lib/cjs/generated/rum.d.ts
```diff
@@ -1740,3 +1744,93 @@ export interface RumRect {
...
+    /**
+     * Indicates whether the system-wide button shapes setting is enabled.
+     */
+    readonly button_shapes_enabled?: boolean;
+    /**
+     * Indicates whether the system-wide color inversion setting is enabled.
+     */
+    readonly invert_colors_enabled?: boolean;
```
mormubis, line 1778: **❓ question:** ‏ Is this meant to be the dark/light mode? If so, I think we should go for a `color-scheme` that let us choose between values.

lib/esm/generated/rum.d.ts
```diff
@@ -1740,3 +1744,93 @@ export interface RumRect {
...
+     */
+    readonly assistive_switch_enabled?: boolean;
+    /**
+     * Indicates whether the system-wide AssistiveTouch feature is currently enabled.
+     */
+    readonly assistive_touch_enabled?: boolean;
+    /**
+     * Indicates whether the video autoplay setting is enabled in the system or application.
+     */
+    readonly is_video_autoplay_enabled?: boolean;
```
mariedm: This is the only property that has the `is_` prefix. Is that on purpose, or should it be removed to align with others?

lib/cjs/generated/rum.d.ts
barboraplasovska: `reduce_motion_enabled` indicates whether the Reduce Motion setting is in an enabled state. ([doc](https://developer.apple.com/documentation/uikit/uiaccessibility/isreducemotionenabled))
`reduce_animations_enabled` indicates whether the Reduce Motion and the Prefer Cross-Fade Transitions settings are in an enabled state. ([doc](https://developer.apple.com/documentation/uikit/uiaccessibility/preferscrossfadetransitions))

lib/cjs/generated/rum.d.ts
barboraplasovska: No. `invert_colors_enabled`  indicates whether the Classic Invert setting is in an enabled state. ([doc](https://developer.apple.com/documentation/uikit/uiaccessibility/isinvertcolorsenabled))

lib/esm/generated/rum.d.ts
barboraplasovska: I will fix it to align with the others, thanks!

lib/cjs/generated/rum.d.ts
mormubis: It seems that `reduce_motion` is only available on iOS, while Android and the web focus more on animations. Should we consolidate this into a single value?

barboraplasovska: Thanks for the suggestion !
Just to give some context, we aligned on these two attributes a while ago through multiple RFCs, and they’re already being sent as custom attributes behind a feature flag. The implementation has been released in the iOS SDK and is currently being dogfooded in the iOS mobile app, where it's also used to populate a dashboard.
At this stage, the work is already quite far along, so I don't think it would make sense to revisit the structure unless there's a critical issue.

mormubis (approved):

mariedm (approved):

Manu-Tran (approved):

PR #271, opened 2025-05-16
👷 add _dd.hidden_while_loading to view events

thomas-lebeau:
The sdk discard loading time if the page was hidden at anytime during its loading. We would like to add this information to the event in order to differentiate it from other instances where loading time are missing for other reason.

Currently I'm suggesting `@view._dd.hidden_while_loading` but as this information could be valuable to the customers to understand why the loading time is missing for a particular view, we could consider make that property public (e.g. `@view.hidden_while_loading`)


rgaignault:
LGTM

rgaignault (approved):

PR #272, opened 2025-06-02
[RUM-10223] Add locale and timezone information to `@device`

mormubis:
# What

This PR adds `locale`, `locales` and `time_zone` to the `device` common field.

# Why

Most of our competitors collect locale and timezone information out-of-the-box. We want to give the same support.

> [!NOTE]
> I needed to remove the requirement from `type` because the browser is not currently sending any device data. Everything was computed in the backend. Now, we want to share only the locale and timezone details.

### DIFF
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -313,9 +313,9 @@
           "description": "The CPU architecture of the device that is reporting the error",
           "readOnly": true
         },
-        "current_locale": {
+        "locale": {
           "type": "string",
-          "description": "The user’s current locale as a language tag combining language and region, e.g. 'en-US'.",
+          "description": "The user’s locale as a language tag combining language and region, e.g. 'en-US'.",
           "readOnly": true
         },
         "locales": {
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -313,6 +313,21 @@
           "type": "string",
           "description": "The CPU architecture of the device that is reporting the error",
           "readOnly": true
+        },
+        "current_locale": {
+          "type": "string",
+          "description": "The user’s current locale as a language tag combining language and region, e.g. 'en-US'.",
+          "readOnly": true
+        },
+        "locales": {
+          "type": "array",
+          "description": "Ordered list of the user’s preferred system languages as IETF language tags.",
+          "readOnly": true
+        },
+        "time_zone": {
+          "type": "string",
+          "description": "The device’s current time zone identifier, e.g. 'Europe/Berlin'.",
+          "readOnly": true
         }
       }
     },
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -286,7 +286,6 @@
     "device": {
       "type": "object",
       "description": "Device properties",
-      "required": ["type"],
       "properties": {
         "type": {
           "type": "string",
--- a/schemas/rum/view-schema.json
+++ b/schemas/rum/view-schema.json
@@ -559,28 +559,6 @@
               "readOnly": true
             }
           }
-        },
-        "device": {
-          "type": "object",
-          "description": "Device properties",
-          "readOnly": true,
-          "properties": {
-            "current_locale": {
-              "type": "string",
-              "description": "The user’s current locale as a language tag combining language and region, e.g. 'en-US'.",
-              "readOnly": true
-            },
-            "locales": {
-              "type": "array",
-              "description": "Ordered list of the user’s preferred system languages as IETF language tags.",
-              "readOnly": true
-            },
-            "time_zone": {
-              "type": "string",
-              "description": "The device’s current time zone identifier, e.g. 'Europe/Berlin'.",
-              "readOnly": true
-            }
-          }
         }
       }
     }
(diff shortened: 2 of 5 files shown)


schemas/rum/view-schema.json
```diff
@@ -397,6 +397,23 @@
...
               "allOf": [{ "$ref": "_view-performance-schema.json" }]
+            },
+            "locale": {
+              "type": "string",
+              "description": "Locale of the user at the time of the view",
+              "readOnly": true
+            },
+            "locales": {
+              "type": "array",
+              "description": "Locale of the user at the time of the view",
```
barboraplasovska: suggestion/ It may be confusing to have both locale and locales with nearly identical names and descriptions. To distinguish them clearly, update the description of locales to:
```"Ordered list of the user’s preferred system languages as IETF language tags."```
This matches the RFC and clarifies that this is a list of preferences, not a single current value.

File: schemas/rum/view-schema.json [hunk 2]
```diff
@@ -397,6 +397,23 @@
             "performance": {
               "description": "Performance data. (Web Vitals, etc.)",
               "allOf": [{ "$ref": "_view-performance-schema.json" }]
+            },
+            "locale": {
+              "type": "string",
+              "description": "Locale of the user at the time of the view",
```
barboraplasovska: suggestion/ Consider renaming this attribute to ```current_locale``` to avoid confusion with the locales array. Also, updating the description to something more precise like:
```"The user’s current locale as an IETF language tag, e.g., 'en-US'."```
This would make the purpose of this field clearer and align better with the proposed naming.

File: schemas/rum/view-schema.json [hunk 3]
```diff
@@ -397,6 +397,23 @@
...
+            "locales": {
+              "type": "array",
+              "description": "Locale of the user at the time of the view",
+              "items": {
+                "type": "string"
+              }
+            },
+            "timezone": {
+              "type": "string",
+              "description": "Timezone of the user at the time of the view",
```
barboraplasovska: suggestion/ For consistency, the description could be more specific: ```"The user’s current time zone identifier, e.g. 'Europe/Berlin'."```

barboraplasovska (commented):
Nice work! 👏 Left some suggestions so that we keep the naming consistent with what was discussed in my RFC.

barboraplasovska (dismissed):
LGTM👌

barboraplasovska (dismissed):

schemas/rum/_common-schema.json
```diff
@@ -313,6 +312,21 @@
           "type": "string",
           "description": "The CPU architecture of the device that is reporting the error",
           "readOnly": true
+        },
+        "current_locale": {
```
thomas-lebeau: **💭 thought:** ‏ sounds a bit weird to have the `current_` prefix, I assume most of these properties are _current_ yet, they don't contain the `current_` prefix (e.g. `timezone`, `connectivity.status`, `display.width`, ...).

thomas-lebeau (commented):
This is a public repository, lets avoid adding links to internal documentation.

schemas/rum/_common-schema.json
mormubis: I agree, but after discussing with some team members we decided to move forward with the current naming

amortemousque: I agree with @thomas-lebeau, current_ is misleading. According to the [HTML spec](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-languages-dev), both `language` and `languages` refer to the user’s `preferred` language(s), not the `current` one.

barboraplasovska (approved):

amortemousque (approved):

RomanWuattier (approved):

Valpertui:
Hello, I'm putting here the comment i've put on the RFC.  

I'm worried this naming would confuse customer inspecting RUM attributes.  
They will see both `locale` and `locales`, and it will not be straightforward why there is two fields without looking at our documentation.
With what we want to implement on mobile, `locales` would not necessarily contain what we find in `locale`.

What about keeping `locale` for the previous "current_locale" field, like the proposal suggest, but also:
- we rename `locales` to `prefered_locales`, which make it clear that it's what the user can/would use.  

Then our customers could use values of `prefered_locale` and `locale` to:
- understand what language they should prioritize for in their translations efforts (You have a lot of user with italian in their prefered locale but your app doesn't support italian)
- evaluate the perceived quality of the translations by the users (you notice Italian users prefers to use the english translation even if your app supports italian)

mormubis:
> Hello, I'm putting here the comment i've put on the RFC.
> 
> I'm worried this naming would confuse customer inspecting RUM attributes. They will see both `locale` and `locales`, and it will not be straightforward why there is two fields without looking at our documentation. With what we want to implement on mobile, `locales` would not necessarily contain what we find in `locale`.
> 
> What about keeping `locale` for the previous "current_locale" field, like the proposal suggest, but also:
> 
> * we rename `locales` to `prefered_locales`, which make it clear that it's what the user can/would use.
> 
> Then our customers could use values of `prefered_locale` and `locale` to:
> 
> * understand what language they should prioritize for in their translations efforts (You have a lot of user with italian in their prefered locale but your app doesn't support italian)
> * evaluate the perceived quality of the translations by the users (you notice Italian users prefers to use the english translation even if your app supports italian)

We kept the conversation offline and agreed to move forward in the meantime

PR #273, opened 2025-06-03
fix: Update `resource._dd.trace_id` regex pattern

ncreated:
## What?

📦 This PR updates the regex pattern for `resource._dd.trace_id` to accept both decimal and hexadecimal formats.

## Why?

Originally, `trace_id` was restricted to decimal values. However, since `RUM-1827` introduced 128-bit trace IDs, we’ve adopted hexadecimal formatting across SDKs and the backend. This update ensures the schema reflects that change.

### DIFF
--- a/schemas/rum/resource-schema.json
+++ b/schemas/rum/resource-schema.json
@@ -326,8 +326,8 @@
             },
             "trace_id": {
               "type": "string",
-              "description": "trace identifier in decimal format",
-              "pattern": "^[0-9]+$",
+              "description": "trace identifier, either a 64 bit decimal number or a 128 bit hexadecimal number padded with 0s",
+              "pattern": "^[0-9]{1,20}|[0-9a-fA-F]{32}$",
               "readOnly": true
             },
             "rule_psr": {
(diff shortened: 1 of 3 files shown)


maciejburda (dismissed):
LGTM

0xnm (dismissed):

schemas/rum/resource-schema.json
```diff
@@ -326,8 +326,8 @@
             },
             "trace_id": {
               "type": "string",
-              "description": "trace identifier in decimal format",
-              "pattern": "^[0-9]+$",
+              "description": "trace identifier in decimal or hexadecimal format",
+              "pattern": "^[0-9a-fA-F]+$",
```
BenoitZugmeyer: **suggestion:** be more precise on the format validation. Only accept a decimal number of max 20 digits or hexadecimal number of 32 digits.

```suggestion
              "description": "trace identifier, either a 64 bit decimal number or a 128 bit hexadecimal number padded with 0s",
              "pattern": "^[0-9]{1,20}|[0-9a-fA-F]{32}$",
```

xgouchet (dismissed):

schemas/rum/resource-schema.json
ncreated: Fair call, updated 👌

BenoitZugmeyer (approved):

xgouchet (approved):

PR #276, opened 2025-06-05
Add view and action tracking to RUM config

mariedm:
This PR updates the telemetry configuration schema with the possible view and action instrumentations options in RUM:
-  ~~UIKit view tracking~~
-  ~~UIKit action tracking~~
- SwiftUI view tracking
- SwiftUI action tracking


BenoitZugmeyer (dismissed):

schemas/telemetry/configuration-schema.json
```diff
@@ -285,6 +285,26 @@
...
+                "uikit_view_tracking_enabled": {
+                  "type": "boolean",
+                  "description": "Whether UIKit view instrumentation is enabled",
+                  "readOnly": false
+                },
+                "uikit_action_tracking_enabled": {
+                  "type": "boolean",
+                  "description": "Whether UIKit action instrumentation is enabled",
+                  "readOnly": false
+                },
```
amortemousque: **❓ question:** ‏ ‏ Out of curiosity, why do you need the prefix? Can both uikit and swiftui instrumentation can be set at the same time?

schemas/telemetry/configuration-schema.json
```diff
@@ -285,6 +285,26 @@
                     "NavigationViewTrackingStrategy"
                   ]
                 },
+                "uikit_view_tracking_enabled": {
```
ncreated: We already track UIKit views and actions configuration through existing schema flags:
- [`track_native_views`](https://github.com/DataDog/dd-sdk-ios/blob/053b4d6b978ae9dba658b8e44a9212d36d1f5a7b/DatadogRUM/Sources/Feature/RUMFeature.swift#L269)
- [`track_user_interactions`](https://github.com/DataDog/dd-sdk-ios/blob/053b4d6b978ae9dba658b8e44a9212d36d1f5a7b/DatadogRUM/Sources/Feature/RUMFeature.swift#L271)

I assume that using it for SwiftUI will cause ambiguity, hence the `uikit_*` prefix proposal, right? What about redundancy?

simaoseica-dd: Having replicated parameters for each one of the UI frameworks doesn't scale. Why not an array of enums like the [track_feature_flags_for_events](https://github.com/DataDog/rum-events-format/blob/master/schemas/telemetry/configuration-schema.json#L434C18-L434C48)?

schemas/telemetry/configuration-schema.json
mariedm: No, these are separate options, which is why we need the prefix. UIKit and SwiftUI are UI frameworks, which have different instrumentations.

schemas/telemetry/configuration-schema.json
mariedm: @ncreated Thanks for pointing that out! I wasn't aware that UIKit instrumentation was already tracked. I'll remove the `uikit_*` fields and only keep the SwiftUI ones.

@simaoseica-dd I agree it would be cleaner, and I would have followed that route if starting from scratch. But since the UIKit configuration is already tracked in the existing schema, I'd rather maintain backward compatibility.

ncreated: 👍 no problem IMO with having less consistency on naming in data format. We can clarify it in the dashboard 👌

ncreated (approved):

BenoitZugmeyer (approved):

PR #279, opened 2025-06-11
add `ddtags` to rum event schema

thomas-lebeau:
adds a few fields the refactor of tags. see https://github.com/DataDog/browser-sdk/pull/3616


BenoitZugmeyer (dismissed):

schemas/rum/_common-schema.json
```diff
@@ -390,6 +395,21 @@
           "type": "string",
           "description": "SDK name (e.g. 'logs', 'rum', 'rum-slim', etc.)",
           "readOnly": true
+        },
+        "retry_count": {
```
mariusc83: don't we send this information also in the request headers ? why do we need it also here ?

thomas-lebeau: Here is the full [RFC](https://docs.google.com/document/d/1UkGnTTsu06w2HFW7sr_RzRMp03gFtQK-pxzygOD1-fo/edit?tab=t.0#heading=h.rnd972k0hiye).
But tl;dr: we need to support sending `ddtags` per log messages (in the json body of a log), and this is ignored if the request url contain `ddtags` too. So we're moving these to the body of the log and the rum event to be consistent

mariusc83: so will this mean that we need to move this information in the RUM events in the SDKs ? This will be quite problematic on our end, I'd say even impossible given the way we are sending/consuming our batches on our end.

thomas-lebeau: This is set as url query parameter `_dd.retry_count=1` instead of `ddtags=retry_count:1`
But as discussed, this is not strictly sent as part of the event, so I'm remove it to the schema

schemas/rum/_common-schema.json
```diff
@@ -44,6 +44,11 @@
       "description": "Generated unique ID of the application build. Unlike version or build_version this field is not meant to be coming from the user, but rather generated by the tooling for each build.",
       "readOnly": true
     },
+    "ddtags": {
+      "type": "string",
+      "description": "tags of the event",
```
ncreated: blocker/ This description is too vague - it is unclear how to format these tags. Since the type is `string`, it raises questions around separators, key-value structure, and formatting expectations. Including a brief example in the description would help clarify usage and improve consistency across SDKs.

ncreated (changes requested):
Looks good overall 👍, but we need more clarity in the description of this field.

BenoitZugmeyer (approved):

ncreated (approved):

PR #282, opened 2025-06-16
RUM-10377 Add battery attributes

barboraplasovska:
Add battery related attributes:
- `battery_level`: Current battery level of the device (0.0 to 1.0)
- `power_saving_mode`: Whether the device is in power saving mode
- `brightness_level`: Current screen brightness level (0.0 to 1.0)

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -1441,6 +1441,18 @@ export interface CommonProperties {
          * The device’s current time zone identifier, e.g. 'Europe/Berlin'.
          */
         readonly time_zone?: string;
+        /**
+         * Current battery level of the device (0.0 to 1.0).
+         */
+        readonly battery_level?: number;
+        /**
+         * Whether the device is in power saving mode.
+         */
+        readonly power_saving_mode?: boolean;
+        /**
+         * Current screen brightness level (0.0 to 1.0).
+         */
+        readonly brightness_level?: number;
         [k: string]: unknown;
     };
     /**
--- a/lib/esm/generated/rum.d.ts
+++ b/lib/esm/generated/rum.d.ts
@@ -1441,6 +1441,18 @@ export interface CommonProperties {
          * The device’s current time zone identifier, e.g. 'Europe/Berlin'.
          */
         readonly time_zone?: string;
+        /**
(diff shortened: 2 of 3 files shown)


lib/cjs/generated/rum.d.ts
```diff
@@ -1003,6 +1003,18 @@ export declare type RumViewEvent = CommonProperties & ViewContainerSchema & {
...
          */
         performance?: ViewPerformanceData;
+        /**
+         * Current battery level of the device (0.0 to 1.0).
+         */
+        readonly battery_level?: number;
+        /**
+         * Whether the device is in low power mode.
+         */
+        readonly low_power_mode?: boolean;
```
maciejburda: nit: this is iOS centric, but maybe we can come up with something shared.

On android it's called Battery Saver if I remember correctly. Not a blocker though.
It's internal model. But on frontend we should make sure we call it in the platform specific way.

maciejburda (dismissed):

schemas/rum/view-schema.json
```diff
@@ -397,6 +397,21 @@
...
+            },
+            "low_power_mode": {
+              "type": "boolean",
+              "description": "Whether the device is in low power mode.",
+              "readOnly": true
+            },
+            "brightness_level": {
+              "type": "number",
+              "description": "Current screen brightness level (0.0 to 1.0).",
+              "readOnly": true
```
mormubis: **❓ question:** ‏ Should these properties be part of the device namespace?

lib/cjs/generated/rum.d.ts
barboraplasovska: I agree, let's try to find an alternative. What do you think about `power_saving_mode` ?

schemas/rum/view-schema.json
barboraplasovska: Good question! I originally added them under `view` because they can change during a session, so collecting them per view made sense for tracking transitions. But I agree that conceptually, they do describe the device state. If others are aligned, I'm happy to move them under the `device` namespace.

mormubis (approved):
LGTM

Manu-Tran (approved):

maciejburda (approved):

maxep (approved):

mariusc83 (approved):

PR #286, opened 2025-06-27
[RUM-10511] Add user story vital type

cy-moi:


lib/cjs/generated/rum.d.ts
```diff
@@ -1146,11 +1146,15 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
...
-        readonly type: 'duration';
+        readonly type: 'duration' | 'user_story';
         /**
          * UUID of the vital
          */
         readonly id: string;
+        /**
+         * UUID for the instance of the User Story allowing to execute the same User Story multiple times in parallel
+         */
+        readonly instance_id?: string;
```
BenoitZugmeyer: **suggestion:** `story_id` or `user_story_id` would be clearer

lib/cjs/generated/rum.d.ts
```diff
@@ -1169,6 +1173,14 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
         readonly custom?: {
             [k: string]: number;
         };
+        /**
+         * Type of the user story event that triggered the vital
+         */
+        readonly event_type?: 'start' | 'touch' | 'retry' | 'end';
```
BenoitZugmeyer: **suggestion:** `step_type` would be better as "event" is already used to depict "RUM events"

lib/cjs/generated/rum.d.ts
```diff
@@ -1146,11 +1146,15 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
         /**
          * Type of the vital
          */
-        readonly type: 'duration';
+        readonly type: 'duration' | 'user_story';
```
BenoitZugmeyer: **suggestion:** the "user_story" will be its own event type eventually:
```suggestion
        readonly type: 'duration' | 'user_story_step';
```

lib/cjs/generated/rum.d.ts
```diff
@@ -1169,6 +1173,14 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
...
             [k: string]: number;
         };
+        /**
+         * Type of the user story event that triggered the vital
+         */
+        readonly step_type?: 'start' | 'touch' | 'retry' | 'end';
+        /**
+         * Reason for the failure of the user story, if applicable
+         */
+        readonly failure_reason?: 'timeout' | 'error' | 'cancelled' | 'other';
```
janakkar: from latest discussions we decided to rename cancelled to abandoned to make a difference for eventual possibility of actually cancelling a user story with the meaning "as if it never happened". Also I don't think we need the timeout? 🤔

guilhermevrs (dismissed):
Looks great, ship it 🚀

lib/cjs/generated/rum.d.ts
cy-moi: We changed this to `parent_id` which indicates that the `step` can be link to a parent.

lib/cjs/generated/rum.d.ts
cy-moi: This should be fixed.

mariedm (approved):
While we wait for the final name, we can start implementing this project! 🚀

guilhermevrs (approved):

kbogtob (approved):

BenoitZugmeyer (approved):

PR #288, opened 2025-07-22
[VQM] define transition schema

eloytoro:
New event type: transtions

These events are part of the QoE data model definition (Read more about it here https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5241110685/QoE+RUM+Payload+Collection)


BeltranBulbarellaDD (dismissed):

lib/cjs/generated/rum.d.ts
```diff
@@ -151,6 +151,50 @@ export declare type RumActionEvent = CommonProperties & ViewContainerSchema & {
...
+         */
+        readonly type: string;
+        /**
+         * UUID of the transition
+         */
+        readonly id?: string;
+        /**
+         * Timestamp in the player
+         */
+        readonly ts?: number;
```
RomanWuattier: Likely a question for the SDK team, should we stay consistent with other events and rename this to `timestamp`?

lib/cjs/generated/rum.d.ts
```diff
@@ -151,6 +151,50 @@ export declare type RumActionEvent = CommonProperties & ViewContainerSchema & {
...
+         */
+        readonly msd?: number;
+        /**
+         * Error code, as reported by the player
+         */
+        readonly ec?: number;
+        /**
+         * Duration of the event in ns
+         */
+        readonly d?: number;
```
RomanWuattier: Likely a question for the SDK team, should we stay consistent with other events and rename this to `duration`?

lib/cjs/generated/rum.d.ts
```diff
@@ -151,6 +151,50 @@ export declare type RumActionEvent = CommonProperties & ViewContainerSchema & {
...
+         */
+        readonly bsd?: number;
+        /**
+         * Media start delay, the amount of time spent loading before playing
+         */
+        readonly msd?: number;
+        /**
+         * Error code, as reported by the player
+         */
+        readonly ec?: number;
```
RomanWuattier: Thinking about how the data will be queried, it may be easier to understand if we avoid using abbreviations. WDYT?

lib/cjs/generated/rum.d.ts
ncreated: +1 for consistency. Also, the description is too ambiguous: what kind of timestamp do we mean (epoch|relative to playback start) and what unit (ns|ms|s)?

lib/cjs/generated/rum.d.ts
```diff
@@ -151,6 +151,50 @@ export declare type RumActionEvent = CommonProperties & ViewContainerSchema & {
...
+         */
+        readonly ts?: number;
+        /**
+         * Buffer starvation duration, the amount of time spent rebuffering
+         */
+        readonly bsd?: number;
+        /**
+         * Media start delay, the amount of time spent loading before playing
+         */
+        readonly msd?: number;
```
ncreated: We should specify units (ns|ms|s), otherwise it will automatically raise questions 🙂.

lib/cjs/generated/rum.d.ts
BeltranBulbarellaDD: +1 for consistency. This is how we have it:
`/**
         * current timestamp at the time of collection
         */
        timestamp?: number;
`

eloytoro: Renamed to `timstamp`

lib/cjs/generated/rum.d.ts
eloytoro: Renamed to `error_code`

lib/cjs/generated/rum.d.ts
eloytoro: added units used to the descriptions

ncreated (dismissed):

BeltranBulbarellaDD (dismissed):

lib/cjs/generated/rum.d.ts
```diff
@@ -151,6 +151,50 @@ export declare type RumActionEvent = CommonProperties & ViewContainerSchema & {
...
+         */
+        readonly id?: string;
+        /**
+         * The player's current timestamp in milliseconds
+         */
+        readonly timestamp?: number;
+        /**
+         * Buffer starvation duration, the amount of time spent rebuffering in milliseconds
+         */
+        readonly buffer_starrvation_duration?: number;
```
RomanWuattier:
```suggestion
        readonly buffer_starvation_duration?: number;
```

RomanWuattier (dismissed):
There is a small typo, but it looks good to me otherwise

barboraplasovska (dismissed):
LGTM👍

schemas/rum/_stream-container-schema.json
```diff
@@ -0,0 +1,51 @@
...
+            }
+          },
+          "readOnly": true
+        },
+        "source": {
+          "type": "string",
+          "description": "Source of the parent stream",
+          "enum": ["android", "ios", "browser", "flutter", "react-native", "roku", "unity", "kotlin-multiplatform"],
+          "readOnly": true
+        }
```
bcaudan: **❓ question:** ‏Since we already have `view.id` and `source` from common schema, do we also need to have them there?

eloytoro: You're right, it's not needed, removing it

bcaudan: What about source?

schemas/rum/_stream-container-schema.json
```diff
@@ -0,0 +1,37 @@
...
+  "type": "object",
+  "description": "Stream Container schema for streams that are nested",
+  "properties": {
+    "container": {
+      "type": "object",
+      "description": "Stream Container properties (stream wrapping the current stream)",
+      "required": ["stream", "source"],
+      "properties": {
+        "stream": {
+          "type": "object",
```
bcaudan: **💬 suggestion:** ‏if source is not needed, we could also remove the container level, no?

schemas/rum/_stream-container-schema.json
eloytoro: yes that one isn't needed either 🤦 removing it...

ncreated (dismissed):

schemas/rum/_stream-container-schema.json
eloytoro: yes, i was basing it off the view-container schema, but that wasn't the correct definition to use as an example

in https://github.com/DataDog/rum-events-format/pull/288/commits/1992a9030a4ab627bec68f5f34770881006ec875 im making it so it's now named `_stream-event-schema.json` and it only makes the `stream.id` property required

bcaudan: Sorry about the iteration on this one, I'm discovering the topic as I am reviewing.
It seems that we would now have a `_stream-event-schema.json` used by view event and a `_stream-schema.json` used by transition event.
Depending on how we plan to use those schemas, we could probably make their names and descriptions more explicit.
Happy to chat about it if needed.

bcaudan (approved):
Nice!

ncreated (approved):

PR #289, opened 2025-07-23
👷 Remove unused vital.custom field in favor of duration vital

cy-moi:
`vital.custom` field is no longer used in browser SDK and never used after GA. For future clarity, we should remove it.


BenoitZugmeyer (dismissed):

BenoitZugmeyer (approved):

mariedm (approved):
Thanks for cleaning this up! 🧹

PR #296, opened 2025-08-20
[Feature Operation] Update `operation_key` definition

mariedm:
`operation_key` is customer-generated and is not a UUID.
 
- Clarifies description
- Corrects type
 
Note: there might be a regex added later, but it's not defined yet.


samcolmanetti (dismissed):
Looks good. Thanks for updating!

satween (dismissed):

cy-moi (approved):

satween (approved):

PR #299, opened 2025-08-28
Add support for allowed GraphQL URLs and payload tracking in telemetry configuration

rgaignault:
See https://github.com/DataDog/browser-sdk/pull/3805

### DIFF
--- a/schemas/telemetry/configuration-schema.json
+++ b/schemas/telemetry/configuration-schema.json
@@ -168,6 +168,10 @@
                   "type": "boolean",
                   "description": "Whether the allowed tracing urls list is used"
                 },
+                "use_allowed_graph_ql_urls": {
+                  "type": "boolean",
+                  "description": "Whether the allowed GraphQL urls list is used"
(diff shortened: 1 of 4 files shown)


schemas/telemetry/configuration-schema.json
```diff
@@ -168,6 +168,10 @@
                   "type": "boolean",
                   "description": "Whether the allowed tracing urls list is used"
                 },
+                "use_allowed_graph_ql_urls": {
```
bcaudan, line 171: **💬 suggestion:** ‏what about also tracking the usage of trackPayload? even if we don't have the detail by endpoint, it could be useful to have an idea of the usage of this option across customers

rgaignault: Indeed it could be interesting, as we discussed also in the rfc the value of it.

bcaudan (approved):

BenoitZugmeyer (approved):

PR #302, opened 2025-09-04
RUM-11595: Extend Vital event to App Launch metrics

simaoseica-dd:
This PR proposes extending the [Vital schema](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/vital-schema.json) to other features. 

Currently, the [Vital schema](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/vital-schema.json) includes some "Feature Operations" parameters. Since the "App Launch" initiative is leveraging vitals, it makes sense to design the schema with scalability in mind.

It introduces new sub schemas of the common `vital-common-schema.json`:
- `vital-duration-schema.json` - Focused on duration properties. It is isolated because Feature Operations doesn't have duration.
- `vital-app-launch-schema.json` - Focused on App launch metrics.
- `vital-feature-operations-schema.json` - Focused on Feature Operations.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -1147,44 +1147,7 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
      * RUM event type
      */
     readonly type: 'vital';
-    /**
-     * Vital properties
-     */
-    readonly vital: {
-        /**
-         * Type of the vital
-         */
-        readonly type: 'duration' | 'operation_step';
-        /**
-         * UUID of the vital
-         */
-        readonly id: string;
-        /**
-         * Optional key to distinguish between multiple operations of the same name running in parallel (e.g., 'photo_upload' with keys 'profile_pic' vs 'cover')
-         */
-        readonly operation_key?: string;
-        /**
-         * Name of the vital, as it is also used as facet path for its value, it must contain only letters, digits, or the characters - _ . @ $
-         */
-        readonly name?: string;
-        /**
-         * Description of the vital. It can be used as a secondary identifier (URL, React component name...)
-         */
-        readonly description?: string;
-        /**
-         * Duration of the vital in nanoseconds
-         */
-        readonly duration?: number;
-        /**
-         * Type of the step that triggered the vital, if applicable
-         */
-        readonly step_type?: 'start' | 'update' | 'retry' | 'end';
-        /**
-         * Reason for the failure of the step, if applicable
-         */
-        readonly failure_reason?: 'error' | 'abandoned' | 'other';
-        [k: string]: unknown;
-    };
+    readonly vital: DurationProperties | AppLaunchProperties | FeatureOperationProperties;
     /**
      * Internal properties
      */
@@ -1201,6 +1164,90 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
         };
         [k: string]: unknown;
     };
+    [k: string]: unknown;
+};
+/**
+ * Duration properties of a Vital event
+ */
+export declare type DurationProperties = VitalCommonProperties & {
+    /**
+     * Type of the vital.
+     */
+    readonly type: 'duration';
+    /**
+     * Duration of the vital in nanoseconds.
+     */
+    readonly duration: number;
+    [k: string]: unknown;
+};
+/**
+ * Schema of common properties for a Vital event
+ */
+export declare type VitalCommonProperties = {
+    /**
+     * UUID of the vital
+     */
+    readonly id: string;
+    /**
+     * Name of the vital, as it is also used as facet path for its value, it must contain only letters, digits, or the characters - _ . @ $
+     */
+    readonly name?: string;
+    /**
+     * Description of the vital. It can be used as a secondary identifier (URL, React component name...)
+     */
+    readonly description?: string;
+    [k: string]: unknown;
+};
+/**
+ * Schema for app launch metrics.
+ */
+export declare type AppLaunchProperties = VitalCommonProperties & {
+    /**
+     * Type of the vital.
+     */
+    readonly type: 'app_launch';
+    /**
+     * The metric of the app launch.
+     */
+    readonly app_launch_metric: 'ttid' | 'ttfd';
+    /**
+     * Duration of the vital in nanoseconds.
+     */
+    readonly duration: number;
+    /**
+     * The type of the app launch.
+     */
+    readonly startup_type?: 'cold_start' | 'warm_start';
+    /**
+     * Whether the app launch was prewarmed.
+     */
+    readonly is_prewarmed?: boolean;
+    /**
+     * If the app launch had a saved instance state bundle.
+     */
+    readonly has_saved_instance_state_bundle?: boolean;
+    [k: string]: unknown;
+};
+/**
+ * Schema for a feature operation.
+ */
+export declare type FeatureOperationProperties = VitalCommonProperties & {
+    /**
+     * Type of the vital.
+     */
+    readonly type: 'operation_step';
+    /**
+     * Optional key to distinguish between multiple operations of the same name running in parallel (e.g., 'photo_upload' with keys 'profile_pic' vs 'cover')
+     */
+    readonly operation_key?: string;
+    /**
+     * Type of the step that triggered the vital, if applicable
+     */
+    readonly step_type?: 'start' | 'update' | 'retry' | 'end';
+    /**
+     * Reason for the failure of the step, if applicable
+     */
+    readonly failure_reason?: 'error' | 'abandoned' | 'other';
     [k: string]: unknown;
 };
 /**
--- a/lib/esm/generated/rum.d.ts
+++ b/lib/esm/generated/rum.d.ts
@@ -1147,44 +1147,7 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
      * RUM event type
      */
     readonly type: 'vital';
-    /**
-     * Vital properties
-     */
-    readonly vital: {
-        /**
-         * Type of the vital
-         */
-        readonly type: 'duration' | 'operation_step';
-        /**
-         * UUID of the vital
-         */
-        readonly id: string;
-        /**
-         * Optional key to distinguish between multiple operations of the same name running in parallel (e.g., 'photo_upload' with keys 'profile_pic' vs 'cover')
-         */
-        readonly operation_key?: string;
-        /**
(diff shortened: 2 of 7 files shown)


lib/esm/generated/rum.d.ts
```diff
@@ -4,7 +4,7 @@
 /**
  * Schema of all properties of a RUM event
  */
-export declare type RumEvent = RumActionEvent | RumErrorEvent | RumLongTaskEvent | RumResourceEvent | RumViewEvent | RumVitalEvent;
+export declare type RumEvent = RumActionEvent | RumErrorEvent | RumLongTaskEvent | RumResourceEvent | RumViewEvent | AppLaunchEvent | FeatureOperationsEvent;
```
aleksandr-gringauz: /question

Is it ok that `RumVitalEvent` is not here anymore? Should you be able to create a vital event with just a duration that is neither `AppLaunchEvent` nor `FeatureOperationsEvent`?

simaoseica-dd: Yes, that does not seem aligned with the current list. Updated.

BenoitZugmeyer (commented):
**Suggestion**: add a `vital-duration-schema.json`, and define a required `duration` property there. I think the `duration` property isn't relevant for feature operation so you could remove it from the "common" schema and declare it where it makes sense

**Suggestion**: move the `vital.type` in their respective schemas and use `oneOf` between schemas. It could focus on the `vital` namespace, and either inherit a "common vital" schema for each type of vital or just do something like:

```js
        "vital": {
          "description": "Vital properties",
          "allOf": [
            {
              "type": "object",
              "required": ["id", "type"],
              "properties": {
                "id": {
                  "type": "string",
                  "description": "UUID of the vital",
                  "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
                  "readOnly": true
                },
                "name": {
                  "type": "string",
                  "description": "Name of the vital, as it is also used as facet path for its value, it must contain only letters, digits, or the characters - _ . @ $",
                  "readOnly": true
                },
                "description": {
                  "type": "string",
                  "description": "Description of the vital. It can be used as a secondary identifier (URL, React component name...)",
                  "readOnly": true
                }
              },
              "readOnly": true
            },
            {
              "oneOf": [
                {
                  "$ref": "vital-duration-schema.json"
                },
                {
                  "$ref": "vital-app-launch-schema.json"
                },
                {
                  "$ref": "vital-feature-operations-schema.json"
                }
              ]
            }
          ],
          "readOnly": true
        },
```

simaoseica-dd:
@BenoitZugmeyer Thank you for the suggestion. The contexts are now truly split with the `oneOf` parameter.

schemas/rum/vital-app-launch-schema.json
```diff
@@ -0,0 +1,45 @@
...
+  "type": "object",
+  "description": "Schema for app launch metrics.",
+  "allOf": [
+    {
+      "required": ["type", "app_launch_time"],
+      "properties": {
+        "type": {
+          "type": "string",
+          "description": "Type of the vital",
+          "const": "duration",
```
BenoitZugmeyer: **suggestion:** 
```suggestion
          "const": "app_launch",
```

schemas/rum/vital-app-launch-schema.json
```diff
@@ -0,0 +1,45 @@
...
+          "readOnly": true
+        },
+        "app_launch_time": {
+          "type": "string",
+          "description": "The time metric of the app launch.",
```
BenoitZugmeyer: **suggestion:** 
```suggestion
          "description": "The time metric of the app launch in nanoseconds.",
```

schemas/rum/vital-common-schema.json
BenoitZugmeyer, line 1: **nitpick:** I would have kept `vital-schema.json` instead of renaming to `vital-common-schema.json`, since other RUM events are named `{event}-schema.json`

schemas/rum/vital-duration-schema.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "rum/vital-duration-schema.json",
+  "title": "DurationProperties",
+  "type": "object",
+  "description": "Duration properties of a Vital event",
+  "required": ["duration"],
+  "properties": {
+    "duration": {
```
BenoitZugmeyer: **suggestion:** the type is missing
```suggestion
  "required": ["type", "duration"],
  "properties": {
    "type": {
      "type": "string",
      "description": "Type of the vital",
      "const": "duration",
      "readOnly": true
    },
    "duration": {
```

schemas/rum/vital-app-launch-schema.json
simaoseica-dd: Initially I was planning to not add more vital types. But I agree with the suggestion, even though we need extra developments on the BE side.

lib/cjs/generated/rum.d.ts
```diff
@@ -1203,6 +1183,54 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
...
+    readonly operation_key?: string;
+    /**
+     * Type of the step that triggered the vital, if applicable
+     */
+    readonly step_type?: 'start' | 'update' | 'retry' | 'end';
+    /**
+     * Reason for the failure of the step, if applicable
+     */
+    readonly failure_reason?: 'error' | 'abandoned' | 'other';
+    [k: string]: unknown;
```
cy-moi, line 1251: *suggestion*: `[k: string]: unknown` should be shared and in `vital-schema`.

lib/cjs/generated/rum.d.ts
```diff
@@ -1151,18 +1151,10 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
      * Vital properties
      */
     readonly vital: {
-        /**
-         * Type of the vital
-         */
-        readonly type: 'duration' | 'operation_step';
```
cy-moi, line 1157: *suggestion*: Keep type as a shared schema, as all sub-schemas require this field.

schemas/rum/vital-app-launch-schema.json
```diff
@@ -0,0 +1,45 @@
...
+          "type": "string",
+          "description": "Type of the vital.",
+          "const": "app_launch",
+          "readOnly": true
+        },
+        "app_launch_time": {
+          "type": "string",
+          "description": "The time metric of the app launch in nanoseconds.",
+          "enum": ["ttid", "ttfd"],
+          "readOnly": true
```
cy-moi: *Question*: The description is confusing here. Is this a time metric (numerical string) or one of `["ttid", "ttfd"]`?

schemas/rum/vital-schema.json
```diff
@@ -21,55 +21,44 @@
...
+            {
+              "oneOf": [
+                {
+                  "$ref": "vital-duration-schema.json"
+                },
+                {
+                  "$ref": "vital-app-launch-schema.json"
+                },
+                {
+                  "$ref": "vital-feature-operations-schema.json"
```
cy-moi: *Nit*: Do we need this to be plural here? I would like to make this singular for easier future reference. Wdyt? cc @mariedm

mariedm: I agree that singular would be better and aligns with the other schema names

schemas/rum/vital-schema.json
```diff
@@ -21,55 +21,44 @@
...
+                  "description": "UUID of the vital",
+                  "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+                  "readOnly": true
+                },
+                "name": {
+                  "type": "string",
+                  "description": "Name of the vital, as it is also used as facet path for its value, it must contain only letters, digits, or the characters - _ . @ $",
+                  "readOnly": true
+                },
+                "description": {
```
mariedm: Will the `AppLaunch` Vital use the `description` property?

I know it's used for the Duration Vital by the Browser SDK. As for FO, we don't use it on mobile SDKs. I believe I already asked @cy-moi, but I don't remember the answer: is the `description` property used on the Browser SDK for FO? If not, we could move it from the shared properties to the `Duration` Vital definition.

mariedm (dismissed):

schemas/rum/vital-schema.json
simaoseica-dd: We're planning to use it. But not as a "secondary identifier" for now

schemas/rum/vital-app-launch-schema.json
simaoseica-dd: Thank you. The `time` was adding confusion so renamed it to `app_launch_metric`.

lib/cjs/generated/rum.d.ts
simaoseica-dd: The vital sub-schema can have other parameters like the `id` that is not defined here.

schemas/rum/vital-app-launch-schema.json
```diff
@@ -0,0 +1,48 @@
...
+        },
+        "has_saved_instance_state_bundle": {
+          "type": "boolean",
+          "description": "If the app launch had a saved instance state bundle.",
+          "readOnly": true
+        }
+      }
+    },
+    {
+      "$ref": "vital-duration-schema.json"
```
aleksandr-gringauz: @simaoseica-dd 

I suggest that we just put the `duration` field directly into this file right next to `is_prewarmed` and others without depending here on `vital-duration-schema.json`.

There is a conflict for the `type` field. This file already has `type = app_launch` and `vital-duration-schema.json` has `type = duration`. They have to be merged somehow. And currently on Android json parser generator fails here.

See how I did it in my pr https://github.com/DataDog/rum-events-format/pull/304/files.

simaoseica-dd: After discussion, we agreed that it was breaking the Json [specification](https://json-schema.org/draft-07/draft-handrews-json-schema-validation-01). Applied your suggestion 👍

cy-moi (approved):

aleksandr-gringauz (approved):
LGTM!

barboraplasovska (approved):

hemeryar (approved):

lib/cjs/generated/rum.d.ts
simaoseica-dd: It is a good point @cy-moi to isolate the sub schemas totally. 
But it will need some extra steps, since it is breaking the "yarn validate"

BenoitZugmeyer (approved):

PR #303, opened 2025-09-05
✨ Stream event schema

mormubis:
Based on the new RFC for Video Quality Monitoring, this implements the schema for stream and adds the ability to add `stream.id` to all events.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -1627,7 +1627,7 @@ export interface StreamSchema {
          */
         bitrate?: number;
         /**
-         * How long is the content (VOD only)
+         * How long is the content (VOD only) (in ms)
          */
         readonly duration?: number;
         /**
@@ -1647,7 +1647,7 @@ export interface StreamSchema {
          */
         timestamp?: number;
         /**
-         * how much did the media progress since the last context update
+         * how much did the media progress since the last context update (in ms)
          */
         watch_time?: number;
         [k: string]: unknown;
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -1533,6 +1533,16 @@ export interface CommonProperties {
     context?: {
         [k: string]: unknown;
     };
+    /**
+     * Stream properties
+     */
+    stream?: {
+        /**
+         * UUID of the stream
+         */
+        readonly id: string;
+        [k: string]: unknown;
+    };
     [k: string]: unknown;
 }
 /**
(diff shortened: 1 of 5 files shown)


lib/cjs/generated/rum.d.ts
```diff
@@ -1604,6 +1614,46 @@ export interface ProfilingInternalContextSchema {
...
+    /**
+     * Stream properties
+     */
+    readonly stream?: {
+        /**
+         * current bitrate at the time of collection
+         */
+        bitrate?: number;
+        /**
+         * How long is the content (VOD only)
```
sbarrio: Perhaps we should explicitly state what unit this duration property uses on this comment (and also on the stream schema below)

lib/esm/generated/rum.d.ts
```diff
@@ -1604,6 +1614,46 @@ export interface ProfilingInternalContextSchema {
...
+         * Stream resolution
+         */
+        readonly resolution?: string;
+        /**
+         * current timestamp at the time of collection
+         */
+        timestamp?: number;
+        /**
+         * how much did the media progress since the last context update
+         */
```
sbarrio, line 1651: I'd also clarify what unit this property is measured in. Is it a percentage?, or elapsed ms?

lib/cjs/generated/rum.d.ts
eloytoro: it should be in milliseconds

eloytoro (approved):

lib/esm/generated/rum.d.ts
mormubis: Added the unit

lib/cjs/generated/rum.d.ts
```diff
@@ -1604,6 +1614,46 @@ export interface ProfilingInternalContextSchema {
...
+         */
+        readonly resolution?: string;
+        /**
+         * current timestamp at the time of collection
+         */
+        timestamp?: number;
+        /**
+         * how much did the media progress since the last context update (in ms)
+         */
+        watch_time?: number;
```
sbarrio, line 1652: (Extreme nitpicking) do we use snake case here for any specific reason?

sbarrio (approved):

lib/cjs/generated/rum.d.ts
mormubis: Added the unit

lib/cjs/generated/rum.d.ts
mormubis: I believe is due to `spent_time` that we use inside the `view`. But I don't have strong opinions about it. cc @eloytoro

cy-moi (approved):

RomanWuattier (approved):

PR #305, opened 2025-09-15
add `trackEarlyRequests` in config telemetry

BenoitZugmeyer:
I want to GA [Early Request Collection](https://github.com/DataDog/browser-sdk/pull/3740), but since it has been evaluated as a small breaking change I want to introduce a `trackEarlyRequests` configuration parameter to opt-in this new behavior.

This PR adds telemetry support for this parameter.

This parameter should be removed in the next major, where we'll enable this new behavior as the default.

### DIFF
--- a/lib/cjs/generated/telemetry.d.ts
+++ b/lib/cjs/generated/telemetry.d.ts
@@ -152,6 +152,10 @@ export declare type TelemetryConfigurationEvent = CommonTelemetryProperties & {
              */
             track_resources?: boolean;
             /**
+             * Whether early requests are tracked
+             */
+            track_early_requests?: boolean;
+            /**
              * Whether long tasks are tracked
              */
             track_long_task?: boolean;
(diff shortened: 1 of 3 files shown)


thomas-lebeau (approved):

lib/cjs/generated/telemetry.d.ts
```diff
@@ -151,6 +151,10 @@ export declare type TelemetryConfigurationEvent = CommonTelemetryProperties & {
              * Whether resources are tracked
              */
             track_resources?: boolean;
+            /**
+             * Whether early requests are tracked
+             */
+            track_early_requests?: boolean;
```
thomas-lebeau, line 157: **💭 thought:** ‏ It's a bit early, but in the long run we might have the same issue with tracking other events (errors, actions, ...)

So do we want to have one option for each event: `track_early_requests`, `track_early_errors`, ...
Or, should we have something like `track_early_data: ['requests', 'errors']` ?

BenoitZugmeyer: In the long run (next major), I want to remove that option and always collect early requests.
For errors, since we want to collect them before `init()`, we can't have an option for it (or we could have an option to just ignore the collected early errors but... I can't imagine why anyone would want that).

thomas-lebeau (approved):

barboraplasovska (approved):

PR #306, opened 2025-09-23
RUM-11840: Add optionality to view on Vital event

simaoseica-dd:
This PR extends the [Vital schema](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/vital-schema.json) to include "[app_launch](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/vital-app-launch-schema.json)" vital requirements: 
- The "app_launch" vitals don't necessarily link to a View. This PR makes that linkage optional.
- The "app_launch" vitals will be associated with profiling data. This PR attaches the [profiling schema](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_profiling-internal-context-schema.json) to the Vital event.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -1216,6 +1216,10 @@ export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
             readonly computed_value?: boolean;
             [k: string]: unknown;
         };
+        /**
+         * Profiling context
+         */
+        profiling?: ProfilingInternalContextSchema;
         [k: string]: unknown;
     };
     [k: string]: unknown;
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -94,7 +94,7 @@ export declare type RumActionEvent = CommonProperties & ViewContainerSchema & {
     /**
      * View properties
      */
-    readonly view?: {
+    readonly view: {
         /**
          * Is the action starting in the foreground (focus in browser)
          */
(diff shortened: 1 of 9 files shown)


lib/cjs/generated/rum.d.ts
```diff
@@ -94,7 +94,7 @@ export declare type RumActionEvent = CommonProperties & ViewContainerSchema & {
     /**
      * View properties
      */
-    readonly view?: {
+    readonly view: {
```
aleksandr-gringauz, line 97: This is interesting...

It was optional before, but now with your change it isn't.

It is ok, but strange nevertheless.

aleksandr-gringauz: I think I get it. Only [this](https://github.com/DataDog/rum-events-format/blob/fbf83da4ab8e7b6ad956aa0038d6e9fef3ec0a9a/schemas/rum/action-schema.json#L135C9-L135C15) part was optional.

aleksandr-gringauz (approved):

lib/cjs/generated/rum.d.ts
simaoseica-dd: Yes, I'm making it a requirement for the `ActionEvent`

schemas/rum/vital-schema.json
Noham-Martin, line 1: Question : since we remove view as "required" in [_common-schema.json](https://github.com/DataDog/rum-events-format/pull/306/files#diff-fffe6f674e634c175a9f7f5a251a3a184711a824958c51eb76baa4e8426c49a7), and don't enforce it in vital-schema, does it means we can send all vital types without view ? Should we require the view in vital-feature-operation-schema.json and vital-duration-schema.json ?

simaoseica-dd: Unfortunately it is not possible since the `view` parameter is defined in the root of the event and these sub schemas are defined in a second level, under `vital`.

BenoitZugmeyer (approved):

Noham-Martin (approved):

PR #308, opened 2025-10-01
Add response error graphql properties

rgaignault:
Add new properties for the following support for graphql response tracking. 

Note : The code property is not native but taken from the extension field and move at the top level. As we would like to query directly : `@resource.[…].errors.code:xxx`. It is present in the spec as a field in the extension part, so it could be manually added by the customers but also some libraries such as Apollo directly provide it in the response. ([see Documentation](https://www.apollographql.com/docs/apollo-server/data/errors))

See the spec for more informations about the graphql response format : [GraphQL Spec](https://github.com/graphql/graphql-spec/blob/main/spec/Section%207%20--%20Response.md)

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -769,6 +769,42 @@ export declare type RumResourceEvent = CommonProperties & ActionChildProperties 
              * String representation of the operation variables
              */
             variables?: string;
+            /**
+             * Number of GraphQL errors in the response
+             */
+            readonly errors_count?: number;
+            /**
+             * Array of GraphQL errors from the response
+             */
+            readonly errors?: {
+                /**
+                 * Error message
+                 */
+                readonly message: string;
+                /**
+                 * Error code from GraphQL extensions
+                 */
+                readonly code?: string;
+                /**
+                 * Array of error locations in the GraphQL query
+                 */
+                readonly locations?: {
+                    /**
+                     * Line number where the error occurred
+                     */
+                    readonly line: number;
+                    /**
+                     * Column number where the error occurred
+                     */
+                    readonly column: number;
+                    [k: string]: unknown;
+                }[];
+                /**
+                 * Path to the field that caused the error
+                 */
+                readonly path?: (string | number)[];
+                [k: string]: unknown;
+            }[];
             [k: string]: unknown;
         };
         [k: string]: unknown;
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -772,7 +772,7 @@ export declare type RumResourceEvent = CommonProperties & ActionChildProperties 
             /**
              * Number of GraphQL errors in the response
              */
-            readonly errors_count?: number;
+            readonly error_count?: number;
             /**
              * Array of GraphQL errors from the response
              */
@@ -782,7 +782,7 @@ export declare type RumResourceEvent = CommonProperties & ActionChildProperties 
                  */
                 readonly message: string;
                 /**
-                 * Error code from GraphQL extensions
+                 * Error code (used by some providers)
                  */
                 readonly code?: string;
                 /**
--- a/samples/rum-events/resource-graphql.json
+++ b/samples/rum-events/resource-graphql.json
@@ -33,7 +33,7 @@
       "operationName": "GetUserProfile",
       "variables": "{\"userId\": \"123\"}",
       "payload": "query GetUserProfile($userId: ID!) { user(id: $userId) { name email } }",
-      "errors_count": 2,
+      "error_count": 2,
       "errors": [
         {
           "message": "User not found",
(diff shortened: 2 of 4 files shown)


lib/cjs/generated/rum.d.ts
```diff
@@ -769,6 +769,42 @@ export declare type RumResourceEvent = CommonProperties & ActionChildProperties
              * String representation of the operation variables
              */
             variables?: string;
+            /**
+             * Number of GraphQL errors in the response
+             */
+            readonly errors_count?: number;
```
BenoitZugmeyer: Suggestion: Error should be singular
```suggestion
            readonly error_count?: number;
```

schemas/rum/resource-schema.json
```diff
@@ -307,6 +307,75 @@
...
+                    "required": ["message"],
+                    "properties": {
+                      "message": {
+                        "type": "string",
+                        "description": "Error message",
+                        "readOnly": true
+                      },
+                      "code": {
+                        "type": "string",
+                        "description": "Error code from GraphQL extensions",
```
bcaudan: **💬 suggestion:** IMO, mentioning GraphQL extensions can be a bit confusing here because it requires to be aware of this concept of the GraphQL spec, and it could easily be misunderstood with the extension concept we have at SDK level.

I think we could:
- Not mention it at all, `"Error code"`
- Not mention GraphQL extension, `"Error code (used by some providers)"`

‏

rgaignault: Sounds good ! I would prefer the second one as it is not always present.  (to inform customers why it's not there)

samples/rum-events/resource-graphql.json
```diff
@@ -0,0 +1,72 @@
...
+        {
+          "message": "User not found",
+          "code": "USER_NOT_FOUND",
+          "locations": [
+            {
+              "line": 1,
+              "column": 15
+            }
+          ],
+          "path": ["user", "name"]
```
bcaudan: **💬 suggestion:** ‏just to have an integer example as well
```suggestion
          "path": ["user", 1]
```

bcaudan (dismissed):

bcaudan (approved):
LGTM

barboraplasovska (approved):
LGTM

satween (approved):

schemas/rum/resource-schema.json
```diff
@@ -307,6 +307,75 @@
...
+                            }
+                          },
+                          "readOnly": true
+                        },
+                        "readOnly": true
+                      },
+                      "path": {
+                        "type": "array",
+                        "description": "Path to the field that caused the error",
+                        "items": {
```
aleksandr-gringauz, line 363: @rgaignault Question.

What does this `path` mean? Is it ok that it is an array of either strings or numbers? Are you sure everything is correct here?

Asking because for Android our json parser generator fails to process this and I want to know if I need to fix it (add support of `oneOf(primitive types)`).

rgaignault: Hey @aleksandr-gringauz ! Based on the spec, that's what I have : https://github.com/graphql/graphql-spec/blob/main/spec/Section%207%20--%20Response.md. 

> The value for a response path must be a list of path segments. Path segments that represent field response name must be strings, and path segments that represent list indices must be 0-indexed integers.

Also based on the examples it's either integer or string.  Do you see an other way of declaring this ?

aleksandr-gringauz: Thanks for the reply @rgaignault!

> Do you see an other way of declaring this ?

For me the following option will work. Checked it in `dd-sdk-android`.

```
"path": {
  "type": "array",
  "description": "Path to the field that caused the error",
  "items": {
    "oneOf": [
      {
        "type": "object",
        "title": "String",
        "required": ["segment"],
        "properties": {
          "segment": {
            "type": "string"
          }
        }
      },
      {
        "type": "object",
        "title": "Integer",
        "required": ["segment"],
        "properties": {
          "segment": {
            "type": "integer"
          }
        }
      }
    ]
  },
  "readOnly": true
}
```

aleksandr-gringauz: The problem with Kotlin is that we don't have `String | Integer` like typescript. We only have `sealed classes`, for example:

```
sealed interface Path {
    data class String(val segment: String): Path
    data class Integer(val segment: Int): Path
}
```

If I were to try to implement `String | Integer` I would have to generate something like:
```
sealed interface StringOrInt {
    data class String(val segment: String): StringOrInt
    data class Integer(val segment: Int): StringOrInt
}
```

And if `oneOf(string|int)` occurs multiple times in the schema, I would have to make different `StringOrInt` for them or somehow reuse a common one.

It all requires some work on the generator which I would like to avoid if possible.

aleksandr-gringauz: Discussed with @maxep about how they do on iOS. I might implement similar solution on Android, but need more time to think.

For now, no action is needed from your side. Thanks!

PR #315, opened 2025-10-22
♻️  Reorganize rum schemas to work around optional "view" property issues

BenoitZugmeyer:
## Motivation

PR https://github.com/DataDog/rum-events-format/pull/306 introduced a change in the rum-events-schemas (making the `view` event property optional for the "App Launch" vital) that is causing some issues within the Browser. 

Even though the `view` property was marked as "required" on events that requires it, this is badly supported by the `json-schema-to-typescript` and the `view` property was still optional on some events (ex: long task, resource...).

On top of this, even if the typescript generation did work correctly, the `RumEvents` union that is exposed to customers via `beforeSend` would still have an optional `view` property since it includes the "App Launch" vital. That would be a breaking change for customers using typescript. In the following example, TypeScript would raise a "'event.view' is possibly 'undefined'." error:

```ts
DD_RUM.init({
  beforeSend(event) {
    event.view.url = removeSensitiveData(url)
  }
})
```

Related Browser SDK PR based on this change: https://github.com/DataDog/browser-sdk/pull/3923

## Changes

To work around this issue, this PR changes how vital schema are declared. Instead of using a single schema for the vital event and a different schema for the `vital` property, it declares one schema for each type of vital event.

This allows to tweak each vital event independently, and make the `view` property optional only on the `App Launch` vital.

This also allows to expose a different `RumEvent` union to be consumed by mobile and browser. For browser, the `RumEvent` union does not include the `App Launch` vital event, so all events have a required `view` property.

Finally, it adds a workaround to make sure the TypeScript generation makes `view` properties required when needed. It adds a validation rule to make sure this workaround is always enforced.

The first two commits are just small maintenance changes related to vital events.


lib/cjs/generated/mobileSessionReplay.d.ts
```diff
@@ -587,6 +559,34 @@ export interface WireframeClip {
...
     readonly right?: number;
 }
+/**
+ * Mobile-specific. Schema of a MutationPayload.
+ */
+export interface MobileMutationPayload {
+    /**
+     * Contains the newly added wireframes.
+     */
+    readonly adds: {
```
rgaignault, line 569: ❓ Question : Why this property is not optional anymore ?

BenoitZugmeyer: Ah good catch. This happened when I did [this change](https://github.com/DataDog/rum-events-format/pull/315/commits/e505ab9f7c984b8fa9ce28e44196ee3eba899888#diff-b5a4061262ff33e9ad221a81d93009e85f27aefe96d94f11d99b11bea49a2ef5). I think the intent was to make it required but it never worked (at least with `json-schema-to-typescript`). Let's wait for a Mobile review to know more.

lib/esm/generated/rum.d.ts
```diff
@@ -1229,119 +1238,133 @@ export declare type RumViewEvent = CommonProperties & ViewContainerSchema & Stre
...
-export declare type RumVitalEvent = CommonProperties & ViewContainerSchema & {
-    /**
-     * RUM event type
-     */
-    readonly type: 'vital';
-    readonly vital: DurationProperties | AppLaunchProperties | FeatureOperationProperties;
+export declare type RumVitalDurationEvent = RumVitalEventCommonProperties & {
+    view: {
+        [k: string]: unknown;
+    };
```
thomas-lebeau, line 1248: **❓ question:** ‏Is it desired to open the view object to any other keys?

I wonder if defining views like this in the different schema will help for this (and for readability?)
```diff
"view": {
-  "type": "object"
+  "$ref": "_common-schema.json#/properties/view"
}
```

thomas-lebeau (commented):
I guess besides the browser SDK, the mobile SDK might also need to update something to account for the mobile/browser schema split.
Any other consumer of these schema we should be aware?

lib/esm/generated/rum.d.ts
BenoitZugmeyer: Interesting! I did the change.

thomas-lebeau (dismissed):

lib/esm/generated/rum.d.ts
BenoitZugmeyer: Although to answer your question: 

> Is it desired to open the view object to any other keys?

[This is already the case](https://github.com/DataDog/rum-events-format/blob/master/lib/esm/generated/rum.d.ts#L1431). We don't use `additionalProperties` so it defaults to `true` everywhere.

Although, `json-schema-to-typescript` has a new option to set the default to `false`, so we could consider getting rid of those `[k: string]: unknown`. Not sure how breaking this would be though 🤔

BenoitZugmeyer:
> I guess besides the browser SDK, the mobile SDK might also need to update something to account for the mobile/browser schema split. Any other consumer of these schema we should be aware?

`web-ui` depends on the generated typescript files, but not the json schema. I'm not aware of other consumers.

rum-events-format.json
maxep, line 1: suggestion/ [We use these entrypoints for our generators](https://github.com/DataDog/dd-sdk-ios/blob/develop/tools/rum-models-generator/run.py#L21-L22), I think it's best to keep them for clarity.

BenoitZugmeyer: Ok let me restore the ones used by your generator.

lib/esm/generated/rum.d.ts
maxep: This breaks our generator, we don't support such notation and it fails to find the ref file :/

aleksandr-gringauz:
I checked the new schema in `dd-sdk-android`.

There is only one problem - https://github.com/DataDog/rum-events-format/pull/308/files#r2454900600

It isn't related to this PR.

The changes from this PR work.

simaoseica-dd (dismissed):
Validated in iOS! 👌 
It makes sense to have different RUM schemas for browser and mobile. ty!

rgaignault (dismissed):

README.md
```diff
@@ -13,19 +13,24 @@ Breaking changes to the format must be reflected by a major version update in `_
...
-Updating to the latest change is possible with _yarn up_, e.g:
+  - Generates TypeScript types from `./schemas/rum-events-browser-schema.json`,
+    `./schemas/telemetry-events-schema.json` and `./schemas/session-replay-browser-schema.json`.
 
-    yarn up rum-events-format@DataDog/rum-events-format#workspace=rum-events-format
+  - Validates data within unit and e2e tests against `./schemas/rum-events-schema.json`.
 
-It is also possible to specify the exact commit hash to use, e.g:
+- The [iOS SDK](https://github.com/DataDog/dd-sdk-ios/) generates native data models based on
+  `./schemas/rum-events-mobile-schema.json`] and `./schemas/session-replay-mobile-schema.json`.
```
thomas-lebeau: **🥜 nitpick:** ‏
```suggestion
  `./schemas/rum-events-mobile-schema.json` and `./schemas/session-replay-mobile-schema.json`.
```

thomas-lebeau (dismissed):

thomas-lebeau (approved):

rgaignault (approved):

PR #326, opened 2025-12-05
Fix validate script + fix new validation errors

BenoitZugmeyer:
The validate script did not iterate correctly on all objects. Also, it looks like `type: object` is not needed for JSON format to consider a type as an object, so we missed some properties validation.

### DIFF
--- a/schemas/rum/_stream-schema.json
+++ b/schemas/rum/_stream-schema.json
@@ -8,7 +8,6 @@
     "stream": {
       "type": "object",
       "description": "Stream properties",
-      "required": ["id"],
       "properties": {
         "bitrate": {
           "type": "number",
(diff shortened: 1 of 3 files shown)


schemas/rum/_stream-schema.json
```diff
@@ -8,7 +8,6 @@
     "stream": {
       "type": "object",
       "description": "Stream properties",
-      "required": ["id"],
```
BenoitZugmeyer, line 11: **note:** this property does not exist

schemas/rum/view-schema.json
```diff
@@ -26,7 +26,7 @@
         "view": {
           "type": "object",
           "description": "View properties",
-          "required": ["id", "url", "time_spent", "action", "error", "resource"],
+          "required": ["time_spent", "action", "error", "resource"],
```
BenoitZugmeyer, line 29: **note:** id and url are required in the parent object (common properties)

BeltranBulbarellaDD (approved):

mormubis (approved):

PR #336, opened 2026-01-07
[PANA-5341] Add constants to make interacting with BrowserChangeRecords easier

sethfowler-datadog:
PR #278 added `BrowserChangeRecord`, a new record type that allows us to express mutations in a more compact way. While that PR included a complete definition of the new additions to the JSON schema, I discovered today that there are some additional type and data structure definitions exposed by `rum-events-format` that are defined directly in TypeScript. This PR updates these definitions to include `BrowserChangeRecord` and its related types. I've imported these definitions from the `browser-sdk` repo; as a followup, I'll remove the duplicated definitions from that repo, so that we can standardize on these ones.


mormubis (approved):

Fhou (approved):
LGTM

PR #337, opened 2026-01-08
RUM-12199: Device Capabilities Attributes Refactoring

marco-saia-datadog:
### What does this PR do?

This PR refines device capability attribute names to improve clarity and reduce ambiguity.

#### `processor_count` → `logical_cpu_count`

The attribute was renamed to explicitly reflect what is actually reported by mobile platform APIs.

* On **Android**, `Runtime.availableProcessors()` returns the number of **logical CPU cores** available to the runtime.
* On **iOS**, `ProcessInfo.processInfo.processorCount` likewise reports the number of **logical** (not physical) cores.

Because both platforms expose logical core counts, the new name and updated description make this behavior explicit and avoid confusion with physical CPU cores.

#### `is_low_ram_device` → `is_low_ram`

The `_device` suffix was removed because the attribute already lives under the `device` parent, making the original name redundant.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -1625,9 +1625,9 @@ export interface CommonProperties {
          */
         readonly brightness_level?: number;
         /**
-         * Number of device processors
+         * Number of logical CPU cores available for scheduling on the device at runtime, as reported by the operating system.
          */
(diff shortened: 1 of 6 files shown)


0xnm (approved):

samcolmanetti (approved):
Looks great!

PR #339, opened 2026-01-16
[RUM-13793] make resource.url optional

BenoitZugmeyer:
On some conditions, the Browser SDK sends resource events without URL:

```js
      DD_RUM.init({
        clientToken: 'xxx',
        applicationId: 'xxx',
        beforeSend() {},
      })
      fetch('/' + 'a'.repeat(300_000))
```

We'll work on a fix, but unfortunately old versions will still be around.

### DIFF
--- a/schemas/rum/resource-schema.json
+++ b/schemas/rum/resource-schema.json
@@ -26,7 +26,7 @@
         "resource": {
           "type": "object",
           "description": "Resource properties",
-          "required": ["type", "url"],
+          "required": ["type"],
           "properties": {
             "id": {
               "type": "string",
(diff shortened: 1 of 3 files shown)


thomas-lebeau (approved):

sethfowler-datadog (approved):

ImaneLargou (approved):

schemas/rum/resource-schema.json
```diff
@@ -26,7 +26,7 @@
         "resource": {
           "type": "object",
           "description": "Resource properties",
-          "required": ["type", "url"],
+          "required": ["type"],
```
0xnm, line 29: Since `resource.url` is nullable, this change will trigger some code changes for languages where nullable type is different from non-nullable.

Will it be a permanent change? Was sending empty string or maybe not sending Resource event at all considered?

BenoitZugmeyer: Let's discuss internally

BenoitZugmeyer:
Resources without URLs will be dropped by the backend.

PR #341, opened 2026-01-25
PANA-5615: Add mobile ui element identifier

jonathanmos:
**Summary**
Adds optional `permanentId` field to mobile session replay common-wireframe-schema and rum-action schemas, for correlating wireframes with RUM action events.

**Changes**
Added `permanentId` to `_common-wireframe-schema.json`
Field: string, optional, readOnly
Description: MD5 hash of element's path (32 lowercase hex characters)

Added `permanentId` to `action-schema.json`
Field: string, optional, readOnly
Description: MD5 hash of element's path (32 lowercase hex characters)

**Affected Types**
All 5 wireframe types inherit this field:
`ShapeWireframe`
`TextWireframe`
`ImageWireframe`
`PlaceholderWireframe`
`WebviewWireframe`

Also, the rum action schema. 

**Compatibility**
Backwards compatible - existing data without `permanentId` remains valid.

### DIFF
--- a/schemas/rum/action-schema.json
+++ b/schemas/rum/action-schema.json
@@ -180,6 +180,11 @@
                       "description": "CSS selector path of the target element",
                       "readOnly": true
                     },
+                    "permanent_id": {
+                      "type": "string",
+                      "description": "Mobile-only: a globally unique and stable identifier for this UI element, computed as the hash of the element's path (32 lowercase hex characters). Used to correlate actions with mobile session replay wireframes.",
+                      "readOnly": true
+                    },
                     "width": {
                       "type": "integer",
                       "description": "Width of the target element (in pixels)",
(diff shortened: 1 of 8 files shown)


nguyexua (approved):

gonzalezreal (approved):
🚀

schemas/rum/action-schema.json
```diff
@@ -180,6 +180,11 @@
                       "description": "CSS selector path of the target element",
                       "readOnly": true
                     },
+                    "component_id": {
+                      "type": "string",
+                      "description": "Mobile-only: a globally unique and stable identifier for this UI element, computed as the hash of the element's path (32 lowercase hex characters). Used to correlate actions with mobile session replay wireframes.",
+                      "readOnly": true
+                    },
```
bcaudan, line 187: **💬 suggestion:** ‏If target and component represent the same concept, we could just use `id` directly.

hugolevyy (approved):

schemas/rum/action-schema.json
jonathanmos: the problem is that this will cause the identifiers in common-wireframe and action-schema to have different names

jonathanmos: following discussion - renamed to `permanentId`

bcaudan (approved):
LGTM

PR #345, opened 2026-02-05
[RUM / Profiling] Add vital and action entries to profiler trace schema

MandragoreVR:
As part of the implementation of profiling aggregation on the backend, we want to include vitals and actions' entries to the profiles' metadata.

This PR updates the schema of the profiler trace to reflect this change.

### DIFF
--- a/lib/esm/generated/browserProfiling.d.ts
+++ b/lib/esm/generated/browserProfiling.d.ts
@@ -135,6 +135,14 @@ export interface BrowserProfilerTrace {
      */
     readonly longTasks: RumProfilerLongTaskEntry[];
     /**
+     * List of detected vital entries.
+     */
+    readonly vitalEntries?: RumProfilerVitalEntry[];
+    /**
+     * List of detected action entries.
+     */
+    readonly actionEntries?: RumProfilerActionEntry[];
+    /**
      * List of detected navigation entries.
      */
     readonly views: RumViewEntry[];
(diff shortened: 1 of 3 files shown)


lib/esm/generated/browserProfiling.d.ts
```diff
@@ -213,7 +221,7 @@ export interface RumProfilerLongTaskEntry {
      */
     readonly id?: string;
     /**
-     * Duration in ns of the long task or long animation frame.
```
MandragoreVR, line 216: This is actually milliseconds, not nanoseconds. The type associated to this [in the browser sdk](https://github.com/DataDog/browser-sdk/blob/main/packages/rum-core/src/domain/longTask/longTaskCollection.ts#L67) is `Duration` which describes milliseconds, and we are directly adding this duration to the start time [in the backend](https://github.com/DataDog/profiling-backend/blob/prod/prof-utils/src/main/java/com/datadog/profiling/aggregation/parser/json/ChromeJsonParser.java#L260C2-L276C3), so this duration is indeed in milliseconds.

BeltranBulbarellaDD: Huh, I was looking at it differently. I aligned it with the unit of time of `RumLongTaskEvent` that is in nanoseconds. [Link](https://github.com/DataDog/browser-sdk/blob/1e1d0fc058daf2f9887f5d025ad02a2845ede5d7/packages/rum-core/src/rumEvent.types.ts#L500C1-L528C72). 
But I won't go against backend 🙂

thomasbertet: Yes what you looked at is the duration that was "processed" (via `toServerDuration`) and in ns, but we are using here the duration directly from the LongAnimationFrame entry that is in ms.

schemas/profiling/browser/profiler-trace-schema.json
```diff
@@ -127,6 +127,60 @@
...
+          "readOnly": true
+        },
+        "label": {
+          "type": "string",
+          "description": "RUM Vital label.",
+          "readOnly": true
+        },
+        "duration": {
+          "type": "number",
+          "description": "Duration in ms of the duration vital.",
```
0xnm: should it be `Duration in ms of the vital`?

schemas/profiling/browser/profiler-trace-schema.json
```diff
@@ -127,6 +127,60 @@
...
+      "required": ["id", "label", "duration", "startClocks"],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "RUM Vital id.",
+          "readOnly": true
+        },
+        "label": {
+          "type": "string",
+          "description": "RUM Vital label.",
```
0xnm, line 143: maybe `name`?

schemas/profiling/browser/profiler-trace-schema.json
```diff
@@ -127,6 +127,60 @@
...
+          "readOnly": true
+        },
+        "label": {
+          "type": "string",
+          "description": "RUM Action label.",
+          "readOnly": true
+        },
+        "duration": {
+          "type": "number",
+          "description": "Duration in ms of the duration vital.",
```
0xnm, line 175:
```suggestion
          "description": "Duration in ms of the vital.",
```

schemas/profiling/browser/profiler-trace-schema.json
MandragoreVR: Yeah, I wanted to underline that we were only selecting duration vitals and not operation steps / app launch vitals, but that's indeed not very clear so I'll remove that.

schemas/profiling/browser/profiler-trace-schema.json
MandragoreVR: `label` is already the field that is [used by the backend](https://github.com/DataDog/profiling-backend/pull/7969/files#diff-fe24f465d52fe19d3a52bfd86ac3eaa5e441e710eca6bb9c26a14cac5ce169e4R443-R448)...

0xnm (approved):

thomasbertet (approved):

BeltranBulbarellaDD (approved):

Noham-Martin (approved):

PR #347, opened 2026-02-11
[PROF] Add action and vital to profiling event schema

MandragoreVR:
This PR is a follow-up of [that one](https://github.com/DataDog/rum-events-format/pull/345), where I forgot to add my new fields `action` and `vital` to the browser `profiler-event-schema`...

### DIFF
--- a/samples/profiling/browser/profile-event/profile-event.json
+++ b/samples/profiling/browser/profile-event/profile-event.json
@@ -9,6 +9,14 @@
   "long_task": {
     "id": ["0e6071e2-e4ae-4605-8dfe-9f3551a172f9"]
   },
+  "action": {
+    "id": ["b3f8c1a2-9d4e-4f5a-8b7c-2e1d3f4a5b6c"],
+    "label": ["action-label"]
+  },
+  "vital": {
+    "id": ["d8205161-044f-482c-975e-77992a55d35f"],
+    "label": ["vital-label"]
+  },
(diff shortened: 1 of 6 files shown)


samples/profiling/browser/profile-event/profile-event.json
```diff
@@ -9,6 +9,14 @@
   "long_task": {
     "id": ["0e6071e2-e4ae-4605-8dfe-9f3551a172f9"]
   },
+  "action": {
+    "id": ["b3f8c1a2-9d4e-4f5a-8b7c-2e1d3f4a5b6c"],
+    "label": ["action-label"]
```
0xnm, line 14: this example shows a bit of inconsistency: `view.name` vs `action.label` / `vital.label` naming

0xnm (approved):

samples/profiling/browser/profile-event/profile-event.json
MandragoreVR: yeah I agree, but we're tied with this name because of [how it's done in the backend](https://github.com/DataDog/profiling-backend/pull/7969)

0xnm: maybe it is not too late to make the alignment on the backend side though (like maybe a simple fallback to another attribute name?)

BeltranBulbarellaDD (approved):

PR #350, opened 2026-02-13
RUM-14055 Add request body size metrics to resource schema

barboraplasovska:
Following the proposal shared on the #rum-rfc Slack channel, this PR introduces the `resource.request` object to the RUM resource event schema with two new fields: `encoded_body_size` (bytes sent over the network after encoding) and `decoded_body_size` (bytes before any encoding).


bcaudan (approved):
Nice 👍

RomanWuattier (approved):

marco-saia-datadog (approved):

PR #354, opened 2026-02-19
Add `rum-cpp` as a source type.

fuzzybinary:
This is for our native C++ SDK.

Additionally, add source_types for native stack trace formats (`pe`,`elf`, and `macho`)

refs: RUM-7416

### DIFF
--- a/lib/cjs/generated/browserProfiling.d.ts
+++ b/lib/cjs/generated/browserProfiling.d.ts
@@ -4,11 +4,11 @@
 /**
  * Schema of Browser SDK Profiling types.
  */
-export type BrowserProfiling = BrowserProfileEvent | BrowserProfilerTrace;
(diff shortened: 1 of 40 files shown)


0xnm (approved):

BenoitZugmeyer (changes requested):
Looks like you're not using the right dependencies versions to build/format stuff. Could you run `yarn` and try again?

Manu-Tran (approved):
We bill by source in the backend, do you need us to mark it as billable or not ? My guess would be not for now ? (see https://github.com/DataDog/logs-backend/blob/7c2adc572a77129ef12f30d4eaf113728d12b74f/domains/rum/libs/rum-commons-domain/src/main/java/com/dd/rum/BillingRumSource.java#L10)

fuzzybinary:
Yes, currently it wouldn't be billable. That PR is coming.

BenoitZugmeyer (approved):

PR #355, opened 2026-02-20
feat: add view_update event type

mormubis:
## Summary

- Add `schemas/rum/view_update-schema.json`: new standalone event schema carrying only changed view fields, with minimal required set (`type`, `application.id`, `session.id`, `view.id`, `_dd.document_version`)
- Add `schemas/rum/_view-properties.json`: shared optional view-specific properties (no `required` at any level) composed by both `view` and `view_update`
- Refactor `schemas/rum/view-schema.json` to reference `_view-properties.json`, removing the large inline property block
- Add `view_update` to `oneOf` in `rum-events-schema.json`, `rum-events-browser-schema.json`, and `rum-events-mobile-schema.json`
- Add `samples/rum-events/view_update.json` sample event
- Regenerate TypeScript types: `RumViewUpdateEvent` now exported in `lib/generated/rum.ts`

## Context

Implements the schema layer for the approved RFC "RUM Event Format limitation" (Oct 2025).

### The problem

Every time anything changes in a user's browser session — scroll depth, error count, time spent on a page — the SDK sends a full view event containing *everything* about that view: user data, feature flags, global context, performance metrics, etc. Even if only one field changed, you get the whole payload. Benchmarks showed **50–90% of that data was redundant**.

### The solution: `view_update`

The new `view_update` event type only carries **what changed**. Instead of resending all 50+ view fields every time, the SDK can send a tiny event:

```json
{
  "type": "view_update",
  "view": { "id": "...", "time_spent": 245512755000 },
  "session": { "id": "..." },
  "application": { "id": "..." },
  "_dd": { "document_version": 2 }
}
```

The backend then merges these updates into the full view state it already has.

### Why `view_update` doesn't extend `_common-schema.json`

The shared `_common-schema.json` requires fields like `session.type` and `view.url`. JSON Schema `allOf` can only *add* required fields, never remove them — so inheriting from the common schema would force those fields to be required on `view_update`, conflicting with the RFC's minimal required set. Instead, `view_update` defines its own properties from scratch.

### The `_view-properties.json` shared schema

All view-specific optional fields (metrics, performance data, feature flags, etc.) were extracted into a shared schema with **no `required` fields at any level**. This lets both `view` and `view_update` reference it safely:
- `view-schema.json` shrank from ~540 lines to ~30 lines
- `view_update-schema.json` composes it alongside its own minimal required fields

## Test plan

- [x] `yarn validate` — all samples pass including `view_update.json`
- [x] `yarn generate` — `RumViewUpdateEvent` exported with `type: "view_update"` discriminator
- [x] `yarn build` — CJS + ESM compile cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)


schemas/rum/view_update-schema.json
```diff
@@ -0,0 +1,501 @@
...
+        },
+        "ddtags": {
+          "type": "string",
+          "description": "Tags of the event in key:value format, separated by commas (e.g. 'env:prod,version:1.2.3')",
+          "readOnly": true
+        },
+        "view": {
+          "type": "object",
+          "description": "View properties",
+          "required": ["id", "url"],
```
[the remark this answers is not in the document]
maxep: Indeed, why `url` is a required field?

schemas/rum/view_update-schema.json
```diff
@@ -0,0 +1,501 @@
...
+    },
+    {
+      "$ref": "_stream-schema.json"
+    },
+    {
+      "$ref": "_view-properties.json"
+    },
+    {
+      "required": ["type", "application", "session", "view", "_dd"],
+      "properties": {
```
maxep, line 22: Looks like we are duplicating the common schema to remove required fields. For maintainability, we should keep a single source for these properties. It would mean making all common properties optional and adding requirements to either each event definition, or to a "required_common" schema.

maxep (commented):
Looks good! The most important feedback is about the common properties duplication in the update event.

schemas/rum/view_update-schema.json
ncreated: ☝️ +1 on this

schemas/rum/view_update-schema.json
```diff
@@ -0,0 +1,46 @@
...
+      "$ref": "_view-container-schema.json"
+    },
+    {
+      "$ref": "_stream-schema.json"
+    },
+    {
+      "$ref": "_view-properties.json"
+    },
+    {
+      "$ref": "_common-schema.json"
```
aleksandr-gringauz, line 18: question.

Do I understand correctly that in the current implementation we will be sending required fields from `common-schema.json` (namely `"date", "application", "session", "_dd"`) in each `ViewUpdate`?

~~Sending `date` every time is redundant, it doesn't change at all since the start of the view.~~ My bad, `date` isn't a good example, it should change actually.

schemas/rum/view-schema.json
```diff
@@ -25,427 +28,45 @@
...
-              "properties": {
-                "count": {
-                  "type": "integer",
-                  "description": "Number of actions that occurred on the view",
-                  "minimum": 0,
-                  "readOnly": true
-                }
-              },
-              "readOnly": true
+              "properties": { "count": { "type": "integer" } }
```
aleksandr-gringauz: is it necessary to duplicate `"properties": { "count": { "type": "integer" } }` here? Will it stop working if we don't do it?

schemas/rum/view_update-schema.json
mormubis: yes! but feel free to suggest otherwise. I chose for this because it's the most straigthforward solution.

schemas/rum/view-schema.json
```diff
@@ -25,427 +28,45 @@
...
-          "description": "Feature flags properties",
-          "additionalProperties": true,
-          "readOnly": true
+          }
         },
         "privacy": {
           "type": "object",
-          "description": "Privacy properties",
           "required": ["replay_level"],
+          "properties": { "replay_level": { "type": "string" } }
```
aleksandr-gringauz: On Android the json schema generator can't process this schema (see https://github.com/DataDog/rum-events-format/actions/runs/22355612083/job/64694296089?pr=355).

```
Execution failed for task ':features:dd-sdk-android-rum:generateRumModelsFromJson'.
> Can't merge Enum with type Primitive(type=STRING, description=, parentType=null)
```

It happens because `replay_level` is a enum in `view-properties.json`, but is a string here. I think you can just remove `"properties": { "replay_level": { "type": "string" } }` here. I checked that it works on android.

schemas/rum/view-schema.json
```diff
@@ -25,427 +28,13 @@
...
-          "properties": {
-            "replay_level": {
-              "type": "string",
-              "description": "The replay privacy level",
-              "enum": ["allow", "mask", "mask-user-input"],
-              "readOnly": true
-            }
-          },
-          "readOnly": true
+            "time_spent": {},
```
aleksandr-gringauz: Android check doesn't pass again :)

It wants `time_spend`, `action`, etc to have `"type": "object"`.

Can we remove these from `view-schema.json` and specify `required` fields in `_view-properties.json` as you already do for other fields?

aleksandr-gringauz: @mormubis 

Now android builds, yes.

But what do you think about specifying the required fields "time_spent", "action", "error", "resource" in `_view-properties.json` instead of `view-schema.json` like I suggest above?

This way we actually keep the definition of the schema inside `_view-properties.json`. Now it looks weird that the required constraints are in the separate file than the definition of the schema itself.

mormubis: I tried, but android kept complaining.

maxep (approved):
Looks great! Tested on the iOS SDK and it builds properly 👍

aleksandr-gringauz (approved):

schemas/rum/_view-properties-schema.json
thomas-lebeau, line 1: **💬 suggestion:** ‏There seems to be an implicit naming convention for the schemas, maybe this file should be called `_view-properties-schema.json` or maybe even `_common-view-properties-schema.json` ?

mormubis: good catch! went with `_view-properties-schema.json` to keep it consistent with the other view-specific schemas in the same dir

thomas-lebeau (approved):

andreicoj (approved):

PR #359, opened 2026-02-25
allow additional properties on @synthetics

BenoitZugmeyer:
Be explicit that we accept additional properties for `@synthetics`. This doesn't change anything in practice since we are accepting additional properties in most places by default.


anediaz (approved):
LGTM

BeltranBulbarellaDD (approved):

PR #361, opened 2026-03-03
Add content_type to resource schema

bdibon:
Add `content_type` to resource schema so we can collect the new `PerformanceResourceTiming.contentType` field.

### DIFF
--- a/schemas/rum/resource-schema.json
+++ b/schemas/rum/resource-schema.json
@@ -244,6 +244,11 @@
               "enum": ["cache", "navigational-prefetch", "other"],
               "readOnly": true
             },
+            "content_type": {
+              "type": "string",
+              "description": "Content type of the resource",
+              "readOnly": true
+            },
             "provider": {
               "type": "object",
               "description": "The provider for this resource",
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -736,6 +736,10 @@ export type RumResourceEvent = CommonProperties & ActionChildProperties & ViewCo
          */
         readonly delivery_type?: 'cache' | 'navigational-prefetch' | 'other';
         /**
+         * Content type of the resource
+         */
+        readonly content_type?: string;
+        /**
          * The provider for this resource
          */
         readonly provider?: {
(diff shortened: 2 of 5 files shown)


schemas/rum/resource-schema.json
```diff
@@ -244,6 +244,11 @@
               "enum": ["cache", "navigational-prefetch", "other"],
               "readOnly": true
             },
+            "content_type": {
+              "type": "string",
+              "description": "Content type of the resource",
+              "readOnly": true
+            },
```
bcaudan, line 251: **💭 thought:** ‏I wonder if we should not provide it as `resource.response.headers.content_type`:
- to avoid potential collision if we want to collect request content type at some point 
- if it can also be collected by the header capture feature

bdibon: This `PerformanceResourceTiming.contentType` field is still experimental, and according to MDN:

> The content type is a minimized and "standardized" version of the MIME type that is extracted from the [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type) HTTP header sent in the resource's fetch response. For JavaScript, JSON, SVG, and XML, the MIME type is replaced by a representative MIME type/subtype string. Other types supported by the browser are represented by the MIME type/subtype string in the header (other information in the header is discarded).

So, if we think about mid / long term, it makes sense to collect the "content-type" information for a given resource in the `http.headers` field.

Plus the Mobile SDKs like Android or iOS plan to implement (or have recently implemented) the collection of some http headers field by default (among which "content-type"), AFAIK they leave it unchanged that is type/subtype and potentially charset and boundary.

On the web we already have the possibility to collect the content-type header as is via the fetch/xhr instrumentation. So, we could do just that and be coherent with what other SDKs are doing.

In a near future, we'll also have access to the aforementioned `PerformanceResourceTiming.contentType` which provides the same information but the eventual charset and boundary will be stripped. Note this also allows us to capture every resource content type that is not retrieved manually via fetch/xhr call.

---

What's best for the product? (these are also questions for myself)
- Having both `http.headers.['content-type']` (i.e `text/javascript; charset=utf-8`) and `resource.content_type` (i.e `text/javascript`)
  - In that case should iOS and Android consider filling this second field too?
  - Is there an impact on cost?
- Having only the `http.headers.['content-type']`, can we easily extract the "text/javascript" piece of information from our UI?

In any case it's easy to fill in the contentType from the performance entry into the http header.

bdibon:
Closing this pull-request, we've decided on using the recently introduced `response.headers` field.

PR #362, opened 2026-03-03
feat(rum): add tab.id to common schema

mormubis:
## Motivation

The Browser SDK shares session state across tabs via cookies/localStorage, but has no way to identify which specific tab produced a given event. Adding `tab.id` enables distinguishing events from different tabs within the same session, enabling tab-level correlation between RUM and Logs.

Reference: [browser-sdk#4184](https://github.com/DataDog/browser-sdk/pull/4184)

## Changes

- Add optional `tab` object with required `id` (UUID) to `_common-schema.json`, at the top level alongside `usr`, `account`, `connectivity`, etc.
- Regenerate TypeScript definitions
- Add `samples/rum-events/view-with-tab.json` as a sample event exercising the new field

## Schema

```json
"tab": {
  "type": "object",
  "description": "Tab properties",
  "required": ["id"],
  "additionalProperties": true,
  "properties": {
    "id": {
      "type": "string",
      "description": "UUID of the browser tab",
      "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
      "readOnly": true
    }
  },
  "readOnly": true
}
```

## Notes

- `tab` is optional at the top level — only browser emits it; existing non-browser events remain valid
- `id` is required when `tab` is present (consistent with `account`)
- `additionalProperties: true` for future enrichment (consistent with `usr` and `account`)

### DIFF
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -176,7 +176,6 @@
       "type": "object",
       "description": "Tab properties",
       "required": ["id"],
-      "additionalProperties": true,
       "properties": {
         "id": {
           "type": "string",
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -174,7 +174,7 @@
     },
     "tab": {
       "type": "object",
-      "description": "Browser tab properties",
+      "description": "Tab properties",
       "required": ["id"],
       "additionalProperties": true,
       "properties": {
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -172,6 +172,21 @@
       },
       "readOnly": true
     },
+    "tab": {
+      "type": "object",
+      "description": "Browser tab properties",
+      "required": ["id"],
+      "additionalProperties": true,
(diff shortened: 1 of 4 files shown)


schemas/rum/_common-schema.json
```diff
@@ -172,6 +172,21 @@
       },
       "readOnly": true
     },
+    "tab": {
+      "type": "object",
+      "description": "Tab properties",
+      "required": ["id"],
+      "additionalProperties": true,
```
BenoitZugmeyer: **nitpick:** Do we need an explicit `additionalProperties: true`? Maybe remove it (it is the default anyway) or make it `false`

BenoitZugmeyer (approved):

schemas/rum/_common-schema.json
mormubis: Removed

sbarrio (approved):
🚀

rodrigol-ddog (approved):
LGTM

PR #370, opened 2026-03-23
[RUM-11993] Add trace context and graphql fields to error event schema

saraSr5:
# Motivation    
  When a network resource fails (timeout, no connectivity), the SDK emits a 'RUMErrorEvent'                                                                                                                                                                                   
  Currently the error event schema has no trace context fields ('trace_id', 'span_id', 'parent_span_id', 'rule_psr')
  and no 'graphql' struct on 'error.resource', unlike the success path ('RUMResourceEvent') which already has both.                                                                                                                                                            
                                                                                                                                                                                                                                                                               
  This means cross-platform SDKs (React Native, Flutter) silently lose trace and GraphQL attributes on the error path.                                                                                                                                                         
                                                                                                                                                                                                                                                                               
  # Changes                                                                                                                                                                                                                                                                                  
  - Added trace context fields ('trace_id', 'span_id', 'parent_span_id', 'rule_psr') to '_dd' in the error event schema, mirroring 'RUMResourceEvent._dd'                                                                                                                      
  - Added 'graphql' struct to 'error.resource', mirroring 'RUMResourceEvent.resource.graphql'
  - Added 'error-schema.json' to the casing exceptions in 'scripts/validate.mjs' (same as 'resource-schema.json', since GraphQL field names follow camelCase convention)

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -800,7 +800,13 @@ export type RumResourceEvent = CommonProperties & ActionChildProperties & ViewCo
     /**
      * Internal properties
      */
-    readonly _dd?: RumTrace;
+    readonly _dd?: RumTrace & {
+        /**
+         * Whether the resource should be discarded or indexed
+         */
+        readonly discarded?: boolean;
+        [k: string]: unknown;
+    };
     [k: string]: unknown;
 };
 /**
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -463,7 +463,13 @@ export type RumErrorEvent = CommonProperties & ActionChildProperties & ViewConta
     /**
      * Internal properties
      */
-    readonly _dd?: RumTrace;
+    readonly _dd?: RumTrace & {
+        /**
+         * Profiling context
+         */
+        profiling?: ProfilingInternalContextSchema;
+        [k: string]: unknown;
+    };
     [k: string]: unknown;
 };
 /**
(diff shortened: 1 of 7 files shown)


lib/cjs/generated/rum.d.ts
```diff
@@ -789,92 +794,13 @@ export type RumResourceEvent = CommonProperties & ActionChildProperties & ViewCo
...
-         * trace sample rate in decimal format
-         */
-        readonly rule_psr?: number;
-        /**
-         * Whether the resource should be discarded or indexed
-         */
-        readonly discarded?: boolean;
-        [k: string]: unknown;
-    };
+    readonly _dd?: RumTrace;
```
mormubis: `RumResourceEvent._dd` is now typed as `RumTrace`, but `RumTrace` doesn't include `discarded`. The field is still in the JSON schema (`resource-schema.json`), so I think the generator is not handling the `allOf` + `properties` composition correctly. I'll check it out.

mormubis (changes requested):

schemas/rum/resource-schema.json
```diff
@@ -433,32 +340,8 @@
...
-            "rule_psr": {
-              "type": "number",
-              "description": "trace sample rate in decimal format",
-              "minimum": 0,
-              "maximum": 1,
-              "readOnly": true
-            },
             "discarded": {
               "type": "boolean",
               "description": "Whether the resource should be discarded or indexed",
```
mormubis: I checked, the generator skips sibling `properties` when it sees `allOf`. Moving `discarded` inside `allOf` as a second item fixes it. I verified locally and the output becomes `RumTrace & { readonly discarded?: boolean }`.


```suggestion
          "allOf": [
            { "$ref": "_trace-schema.json" },
            {
              "properties": {
                "discarded": {
                  "type": "boolean",
                  "description": "Whether the resource should be discarded or indexed",
                  "readOnly": true
                }
              }
            }
          ],
```
Note: you cannot merge the suggestion as it doesn't show the full discarded field

saraSr5: Okey thank you I'll take a look and change it!

mormubis (approved):

maxep (approved):
LGTM 👍

hugolevyy (approved):

PR #373, opened 2026-03-27
[PANA-6704] Allow multiple formats in BrowserFullSnapshot records

sethfowler-datadog:
For better compatibility with older mobile SDKs, it would be helpful if `BrowserChange` records were represented as `BrowserFullSnapshot` records when they represented a full snapshot. This would enable mobile SDKs which receive these records from webviews to set metadata accurately on the segments that include these records.

This PR makes this possible by adding an optional `format` property to `BrowserFullSnapshot` records. When `format` is unset, or has the value `SnapshotFormatV1`, the `data` property is interpreted as containing a snapshot in original V1 format. When `format` has the value `SnapshotFormatChange`, the `data` property is interpreted as containing a snapshot in the new Change format.

This change is backwards compatible from a data format perspective, since any `BrowserFullSnapshot` without a `format` property is interpreted in the same way it would be today, but code which interacts with the `data` property may need changes to avoid triggering TypeScript errors. I'll make the necessary changes in downstream repos after this lands.


simaoseica-dd (approved):
LGTM! And no ramifications to Mobile. 👌

mormubis (approved):
LGTM!

PR #374, opened 2026-03-31
[RUM-15273] Add telemetry for trackResourceHeaders

bdibon:
## Summary

Add optional `use_track_resource_headers` to the telemetry **configuration** event schema (`telemetry/configuration-schema.json`).

- **Type:** string enum: `default_headers` | `custom`
- **Semantics (Browser RUM SDK):**
  - Omitted when `trackResourceHeaders` is `undefined` or `false`
  - `default_headers` when the option is `true` (built-in default header set)
  - `custom` when the option is an array (including empty arrays)

No header names, URLs, or patterns are sent — only how the option was set at init.

## Test plan

- [x] `yarn validate` passes
- [x] `yarn generate` produces updated types
- [x] `yarn format` reports no issues

### DIFF
--- a/lib/cjs/generated/telemetry.d.ts
+++ b/lib/cjs/generated/telemetry.d.ts
@@ -462,6 +462,10 @@ export type TelemetryConfigurationEvent = CommonTelemetryProperties & {
              */
             propagate_trace_baggage?: boolean;
             /**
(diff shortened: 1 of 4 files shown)


kikoveiga (approved):

bcaudan (approved):
LGTM

PR #379, opened 2026-04-16
[Android only] RUM-15682: Adding legacy okhttp option

satween:
Adding `LEGACY_OKHTTP` field for networking api usage telemetry on android.

We are migrating to the new infrastructure for the network instrumentation. 
This telemetry would help us to see the adoption rate between old (LEGACY_OKHTTP) and new (OKHTTP) infrastructure and CRONET

### DIFF
--- a/schemas/telemetry/usage/mobile-features-schema.json
+++ b/schemas/telemetry/usage/mobile-features-schema.json
@@ -28,7 +28,7 @@
         "type": {
           "type": "string",
           "description": "The network instrumentation API used",
-          "enum": ["CRONET", "OKHTTP"]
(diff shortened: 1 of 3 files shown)


schemas/telemetry/usage/mobile-features-schema.json
```diff
@@ -28,7 +28,7 @@
         "type": {
           "type": "string",
           "description": "The network instrumentation API used",
```
0xnm, line 30: nit: it is worth maybe mentioning in the description what is the difference between `OKHTTP` and `LEGACY_OKHTTP`

0xnm (approved):

kikoveiga (approved):

hamorillo (approved):

PR #387, opened 2026-05-11
RUM-16231: Create schema for RUM events metadata sent with profiling events on Mobile SDKs

0xnm:
This PR creates schema for the RUM events metadata sent as a part of the profiling event upload in a dedicated file.

It is used to associate profiling samples with the timeframe of the particular RUM event represented by the start timestamp and duration (optional, can be missing if event is still ongoing by the time profile event ends).

### DIFF
--- a/schemas/profiling-mobile-rum-metadata-event-schema.json
+++ b/schemas/profiling-mobile-rum-metadata-event-schema.json
@@ -1,0 +1,7 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "profiling-mobile-rum-metadata-event-schema.json",
+  "title": "MobileProfileRumMetadataEvent",
+  "description": "Schema of the Mobile SDK RUM Metadata event(s) attached as auxiliary array to Profile Event.",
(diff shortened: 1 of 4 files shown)


mormubis (approved):

schemas/profiling-mobile-rum-metadata-event-schema.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "profiling-mobile-rum-metadata-event-schema.json",
+  "title": "MobileProfileRumMetadataEvent",
+  "description": "Schema of the Mobile SDK RUM Metadata event(s) attached as auxiliary array to Profile Event.",
+  "$ref": "profiling/mobile/profile-rum-metadata-event-schema.json"
```
ambushwork, line 6: Question, this is referring an object type, how it can be an array?

0xnm: Array is created manually, schema doesn't represent the array, but individual events. The same way as we do basically for RUM events (we don't declare the whole array there).

ambushwork (approved):

PR #390, opened 2026-05-29
[PANA-7586] Add composition tree schema for Mobile Session Replay

gonzalezreal:
Mobile Session Replay wireframes are flat, which means the player has no way to apply rendering operations that depend on hierarchy (clipping with rounded corners, group opacity, blur, and compositing). This PR adds an optional `compositionTree` to mobile full snapshots that lets SDKs describe the rendering groups the player needs.

### What's new

A composition layer groups child wireframes and child layers and carries an ordered list of rendering modifiers: `clip`, `opacity`, `colorMatrix`, `gaussianBlur`, `brightnessBias`, `saturate`, and `backgroundMaterial`. Each layer can also specify a `compositeOperation` for blending into its parent.

A new incremental data source (`source: 10`) carries layer mutations following the same model as wireframe mutations.

### Backward compatibility

`compositionTree` is optional. When it is absent, the player renders the flat `wireframes` array as before.

Internal references:
- [Composition tree schema RFC](https://datadoghq.atlassian.net/wiki/x/tIVXkQE)
- [Core Animation recording RFC](https://datadoghq.atlassian.net/wiki/x/v4AXjAE)

### DIFF
--- a/schemas/session-replay/mobile/composition-layer-update-schema.json
+++ b/schemas/session-replay/mobile/composition-layer-update-schema.json
@@ -3,7 +3,7 @@
   "$id": "session-replay/mobile/composition-layer-update-schema.json",
   "title": "CompositionLayerUpdate",
   "type": "object",
-  "description": "Sparse update for a composition layer. Only id is required; omitted fields are unchanged. An empty modifiers array clears all modifiers. An empty children array clears all children.",
+  "description": "Sparse update for a composition layer. Omitted fields are unchanged.",
   "required": ["id"],
   "properties": {
     "id": {
@@ -44,7 +44,7 @@
       "items": {
         "$ref": "composition-layer-modifier-schema.json"
       },
-      "description": "When present, replaces the full modifier list for this layer. An empty array clears all modifiers.",
+      "description": "When present, replaces the full modifier list for this layer.",
       "readOnly": true
     },
     "compositeOperation": {
--- a/schemas/session-replay/mobile/composition-layer-update-schema.json
+++ b/schemas/session-replay/mobile/composition-layer-update-schema.json
@@ -1,0 +1,57 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/mobile/composition-layer-update-schema.json",
+  "title": "CompositionLayerUpdate",
+  "type": "object",
+  "description": "Sparse update for a composition layer. Only id is required; omitted fields are unchanged. An empty modifiers array clears all modifiers. An empty children array clears all children.",
+  "required": ["id"],
+  "properties": {
+    "id": {
+      "type": "integer",
+      "description": "The id of the layer to update.",
+      "readOnly": true
+    },
+    "x": {
+      "type": "integer",
+      "description": "Updated X position in absolute coordinates. Uses the same coordinate space as mobile wireframes.",
+      "readOnly": true
+    },
+    "y": {
+      "type": "integer",
+      "description": "Updated Y position in absolute coordinates. Uses the same coordinate space as mobile wireframes.",
+      "readOnly": true
+    },
+    "width": {
+      "type": "integer",
+      "description": "Updated width in pixels. Uses the same coordinate space as mobile wireframes.",
+      "readOnly": true
+    },
+    "height": {
+      "type": "integer",
+      "description": "Updated height in pixels. Uses the same coordinate space as mobile wireframes.",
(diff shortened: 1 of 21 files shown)


schemas/session-replay/mobile/composition-layer-update-schema.json
```diff
@@ -0,0 +1,57 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/mobile/composition-layer-update-schema.json",
+  "title": "CompositionLayerUpdate",
+  "type": "object",
+  "description": "Sparse update for a composition layer. Only id is required; omitted fields are unchanged. An empty modifiers array clears all modifiers. An empty children array clears all children.",
```
sethfowler-datadog: Nit: Some of this description feels a bit redundant with the descriptions of the individual fields; maybe just this would be better:

```suggestion
  "description": "Sparse update for a composition layer. Omitted fields are unchanged.",
```

schemas/session-replay/mobile/composition-layer-update-schema.json
```diff
@@ -0,0 +1,57 @@
...
+      },
+      "description": "When present, replaces the full child list for this layer.",
+      "readOnly": true
+    },
+    "modifiers": {
+      "type": "array",
+      "items": {
+        "$ref": "composition-layer-modifier-schema.json"
+      },
+      "description": "When present, replaces the full modifier list for this layer. An empty array clears all modifiers.",
```
sethfowler-datadog: Nit: "An empty array clears all modifiers." is implied by "When present, replaces the full modifier list for this layer.", I think. It's potentially confusing that the extra explanation is present here, but not for `children`, which might lead a reader to think that the two have different behavior. I'd suggest:

```suggestion
      "description": "When present, replaces the full modifier list for this layer.",
```

schemas/session-replay/mobile/composition-tree-mutation-data-schema.json
```diff
@@ -0,0 +1,43 @@
...
+      "readOnly": true
+    },
+    "removes": {
+      "type": "array",
+      "items": {
+        "type": "integer"
+      },
+      "description": "Ids of layers to remove.",
+      "readOnly": true
+    },
```
sethfowler-datadog, line 33: In this schema you have two ways of expressing the removal of a layer:
1. You can send the layer's id in `removes`.
2. You can remove the layer's id from the list of children of its parent in `updates`.

Is a correct implementation required to use both approaches?

Would it make sense to eliminate this redundancy in the schema? (In the equivalent schemas for web, we don't have this redundancy, because we have an "adds" operation for mutating child lists along with "removes", and we don't have a "replace the entire child list" operation.)

schemas/session-replay/mobile/composition-tree-schema.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "session-replay/mobile/composition-tree-schema.json",
+  "title": "CompositionTree",
+  "type": "object",
+  "description": "Optional composition tree describing the rendering hierarchy for a full snapshot. When present, the player uses this tree for rendering order and group operations instead of rendering the flat wireframes array.",
```
sethfowler-datadog: Nit:

```suggestion
  "description": "Optional composition tree describing the rendering hierarchy for a full snapshot. When present, the player uses this tree for rendering order and group operations instead of rendering the wireframes as a flat array.",
```

sethfowler-datadog (commented):
Everything generally looks good! I had one potentially-blocking question about child list updates, though.

schemas/session-replay/mobile/composition-tree-mutation-data-schema.json
gonzalezreal: IMO they are not equivalent operations.

Updating `children` changes the tree relationship. `removes` deletes layer definitions from the composition layer store. If a layer is currently referenced by a parent, a complete removal should remove the reference from the parent / root child list and include the layer id in `removes`.

I kept full child-list replacement because it matches the mobile SDK diffing shape better: the SDK can compare the previous and current layer snapshot and replace the affected child list without modeling insert/remove child operations. The tradeoff is that deleting a referenced layer has two parts in the normalized representation.

I updated the `removes` description to make this explicit.

mormubis (approved):

sethfowler-datadog (approved):

PR #396, opened 2026-06-19
Add missing _dd fields to view_update schema

mormubis:
## Motivation

The `view_update` schema was introduced with only `document_version` in `_dd`. But several other `_dd` fields from the full view schema can change during a view's lifecycle — `page_states` as the user navigates, `replay_stats` as replay records accumulate, `cls` when layout shifts are detected, and `configuration` when it needs to be reported at update time. Without these in the schema, the SDK can't legally include them in a view update event.

## Changes

Added `page_states`, `replay_stats`, `cls`, and `configuration` to `_dd` in `view_update-schema.json` as optional fields, with definitions copied from `view-schema.json`. `document_version` stays the only required field.

Regenerated TypeScript types.


maxep (commented):
Thanks @mormubis 🙏 What about `_dd.profiling`?

mormubis:
> Thanks @mormubis 🙏 What about `_dd.profiling`?

@maxep Updated it!

maxep (approved):
🙏

hamorillo (approved):
🚀

bcaudan (approved):

andreicoj (approved):

PR #406, opened 2026-07-09
Update timeseries CPU and memory schemas to object-v2 flattened format

barboraplasovska:
### What and why?

- `timeseries.schema` is now a const `"object-v2"` (removes `object`/`delta-scalar`/`delta-object` enum)
- `timeseries.data` flattened from array of `{timestamp, data_point}` to `{ timestamps: [], values: {...} }`
- `memory_footprint` unit changed from bytes to KB (still float)

### Review checklist
- [ ] Schema changes reviewed for shape correctness
- [ ] No breaking change to `id`/`start`/`end`/`name` fields

### DIFF
--- a/schemas/rum/timeseries-memory-schema.json
+++ b/schemas/rum/timeseries-memory-schema.json
@@ -111,7 +111,7 @@
         "schema": {
           "type": "string",
           "description": "Wire-shape discriminator for the data field",
-          "enum": ["object", "delta-object"],
+          "const": "object-v2",
           "readOnly": true
         },
         "start": {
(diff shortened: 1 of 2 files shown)


schemas/rum/timeseries-memory-schema.json
```diff
@@ -125,38 +125,45 @@
...
-            "required": ["timestamp", "data_point"],
-            "properties": {
-              "timestamp": {
+          "type": "object",
+          "description": "Flattened memory data points",
+          "required": ["timestamps", "values"],
+          "properties": {
+            "timestamps": {
+              "type": "array",
+              "description": "Sample timestamps in nanoseconds from epoch",
```
satween, line 134: I could be wrong, but aren't we decided to use milliseconds?

barboraplasovska: No, we're keeping nanoseconds for now.

wlmsrvty (approved):

satween (approved):

PR #412, opened 2026-07-16
RUM-17207 Add local_cache_hit field to resource event schema

barboraplasovska:
### What and why?

Adds an optional, read-only `local_cache_hit` boolean field to the Resource so mobile SDKs can report when a resource was served from the device's local cache. The RUM backend has accepted `resource.local_cache_hit == true` as an input signal since 15/06 (to set `cache_reason = local_cache`), but no SDK can send it yet since the field doesn't exist in the schema.

### DIFF
--- a/schemas/rum/resource-schema.json
+++ b/schemas/rum/resource-schema.json
@@ -244,6 +244,11 @@
               "enum": ["cache", "navigational-prefetch", "other"],
               "readOnly": true
             },
+            "local_cache_hit": {
+              "type": "boolean",
+              "description": "Mobile-only: indicates the resource was served from the device's local cache. Only ever sent as `true`; absence means not a local cache hit.",
+              "readOnly": true
+            },
             "provider": {
               "type": "object",
               "description": "The provider for this resource",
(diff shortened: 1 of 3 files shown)


simaoseica-dd (approved):

schemas/rum/resource-schema.json
samcolmanetti, line 241: any reason why we can't reuse `@resource.delivery_type`? browsers currently indicate local cache through this and is not platform specific. 

cc @qsellem as author of the product brief

schemas/rum/resource-schema.json
```diff
@@ -244,6 +244,11 @@
               "enum": ["cache", "navigational-prefetch", "other"],
               "readOnly": true
             },
+            "local_cache_hit": {
```
mormubis, line 247: Edit: removed my comment, I just saw @samcolmanetti comment

kikoveiga (approved):

schemas/rum/resource-schema.json
samcolmanetti: the team discussed over slack. settled on local_cache_hit and a future server_validated_cache 👍

samcolmanetti (approved):

PR #415, opened 2026-07-21
⚗️ [RUM-17561] add execution_context event type to RUM schema

bcaudan:
## Motivation

Introduces a first-class `execution_context` event type to support Electron child-process monitoring. Allows the SDK to report execution context lifecycle data (start time, duration, exit reason) and links all existing RUM events back to their owning execution context via optional enrichment on the common schema.

## Changes

- New `execution_context-schema.json` event type extending `_common-schema.json`. Required fields: `type` (constant `"execution_context"`), `execution_context.id` (UUID), `execution_context.type` (enum: `main-process`/`renderer-process`/`utility-process`, Electron-only for now), `execution_context.instance_id`, `_dd.document_version`. Optional: `execution_context.parent_instance_id`, `execution_context.name`, `execution_context.duration` (ns), `execution_context.exit_reason`.
- `execution_context-schema.json` added to `rum-events-electron-schema.json` and the generic `rum-events-schema.json` union.
- Optional `execution_context.id`, `execution_context.type`, and `execution_context.name` enrichment added to `_common-schema.json` so all existing event types (Error, Action, Resource, etc.) can carry a back-reference to their owning execution context.
- Optional `view.is_fake` boolean added to `_common-schema.json` so all event types associated with a fake view can carry the flag for backend and frontend filtering. Fake views are synthetically created by the SDK to carry execution context events during the transition phase where execution context events require a `view.id`.

### DIFF
--- a/schemas/rum-events-schema.json
+++ b/schemas/rum-events-schema.json
@@ -27,6 +27,9 @@
       "$ref": "rum/view_update-schema.json"
     },
     {
+      "$ref": "rum/process-schema.json"
+    },
+    {
       "title": "RumVitalEvent",
       "oneOf": [
         {
--- a/schemas/rum-events-schema.json
+++ b/schemas/rum-events-schema.json
@@ -27,7 +27,7 @@
       "$ref": "rum/view_update-schema.json"
     },
     {
-      "$ref": "rum/process-schema.json"
+      "$ref": "rum/execution_context-schema.json"
     },
     {
       "title": "RumVitalEvent",
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -122,6 +122,11 @@
           "type": "string",
           "description": "User defined name of the view",
           "readOnly": false
+        },
+        "is_fake": {
+          "type": "boolean",
+          "description": "Whether this view was synthetically created to carry view-less events",
+          "readOnly": true
         }
       },
       "readOnly": true
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -510,6 +510,25 @@
           "readOnly": true
         }
       }
+    },
+    "process": {
+      "type": "object",
+      "description": "Process properties",
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "UUID of the process",
+          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+          "readOnly": true
+        },
+        "role": {
+          "type": "string",
+          "description": "Role of the process",
+          "enum": ["main", "renderer", "utility"],
+          "readOnly": true
+        }
+      },
+      "readOnly": true
     }
   }
 }
--- a/schemas/rum/_common-schema.json
+++ b/schemas/rum/_common-schema.json
@@ -516,25 +516,25 @@
         }
       }
     },
-    "process": {
+    "execution_context": {
       "type": "object",
-      "description": "Process properties",
+      "description": "Execution context properties",
       "properties": {
         "id": {
           "type": "string",
-          "description": "UUID of the process",
+          "description": "UUID of the execution context",
           "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
           "readOnly": true
         },
         "role": {
           "type": "string",
-          "description": "Role of the process",
+          "description": "Role of the execution context",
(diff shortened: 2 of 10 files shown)


schemas/rum/execution_context-schema.json
```diff
@@ -0,0 +1,81 @@
...
+              "type": "string",
+              "description": "Process name",
+              "readOnly": true
+            },
+            "duration": {
+              "type": "integer",
+              "description": "Process lifetime in nanoseconds",
+              "minimum": 0,
+              "readOnly": true
+            },
```
bcaudan, line 57: **💭 thought:** ‏Used `date` / `process.duration` instead of `process.start` / `process.end` to be consistent with other events.

schemas/rum/_view-properties-schema.json
```diff
@@ -556,6 +556,12 @@
           "type": "object",
           "description": "Profiling context",
           "allOf": [{ "$ref": "./_profiling-internal-context-schema.json" }]
+        },
+        "is_fake": {
+          "type": "boolean",
+          "const": true,
+          "description": "Whether this view was synthetically created to carry view-less events",
+          "readOnly": true
```
bcaudan: **💭 thought:** ‏I put it in `_dd.is_fake` so we could use the same path for fake process events as well.

schemas/rum/_common-schema.json
```diff
@@ -502,6 +502,25 @@
...
+      "type": "object",
+      "description": "Process properties",
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "UUID of the process",
+          "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$",
+          "readOnly": true
+        },
+        "role": {
```
Marsonge: Shouldn't this only live in the `process` event? 🤔

bcaudan: It is also there, I've added it here as it maybe handy to display the process role on the event hierarchy.
I think we should probably also denormalize the process name here.

bcaudan: added in the commit 👇

schemas/rum/_view-properties-schema.json
bcaudan: Move to `view.is_fake`, since we won't use fake for process and it will be used on child events too.
So let's colocate in the view namespace.

Marsonge (approved):

schemas/rum-events-schema.json
```diff
@@ -26,6 +26,9 @@
     {
       "$ref": "rum/view_update-schema.json"
     },
+    {
+      "$ref": "rum/execution_context-schema.json"
```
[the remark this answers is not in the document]
bcaudan: Added a new electron branch to separate from browser `RUMEvent` types, then we could add the new type to browser and mobile schemas when they'll need to support it.

schemas/rum/execution_context-schema.json
```diff
@@ -0,0 +1,93 @@
...
+  "$id": "rum/execution_context-schema.json",
+  "title": "RumExecutionContextEvent",
+  "type": "object",
+  "description": "Schema of all properties of an Execution Context event",
+  "allOf": [
+    {
+      "$ref": "_common-schema.json"
+    },
+    {
+      "required": ["type", "execution_context", "_dd"],
```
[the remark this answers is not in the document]
bcaudan: Let's keep it optional to avoid adding a required constraint that we plan to relax.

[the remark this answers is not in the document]
bcaudan: Let's keep it optional to avoid adding a required constraint that we plan to relax.

bcaudan:
@autotest review

schemas/rum/execution_context-schema.json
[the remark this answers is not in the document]
bcaudan: Let's keep it optional to avoid adding a required constraint that we plan to relax.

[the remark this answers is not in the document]
bcaudan: Let's keep it optional to avoid adding a required constraint that we plan to relax.

schemas/rum/_common-schema.json
```diff
@@ -510,6 +515,30 @@
           "readOnly": true
         }
       }
+    },
+    "execution_context": {
```
rgaignault, line 519: ❓ Question: Don’t we need to require id when `execution_context` is present, since it is what associates the event with its execution context?

rgaignault (approved):

schemas/rum/_common-schema.json
bcaudan: Nice catch, I'll add that 👍

PR #416, opened 2026-07-22
Juan.naranjo/flutter embedded view schema

JuanNaranjoDD:
**What and why?**

Reorders the oneOf entries for embedded-content-wireframe(-update)-schema.json in wireframe-schema.json and wireframe-update-mutation-schema.json, moving them to the front of the list instead of the end.

Generated Swift/Kotlin decoders try each oneOf case in declaration order and keep the first one that successfully decodes. SRShapeWireframe's required fields are a strict subset of SREmbeddedContentWireframe's, so with embedded-content listed last, an embedded-content payload was being matched and decoded as a generic shape wireframe first — silently dropping its required slotId and corrupting the type field. This broke any round-trip decode of an embedded-content wireframe (e.g. Flutter add-to-app placeholders), even though encoding was always correct.

**How?**

Moved the embedded-content-wireframe-schema.json ref to the first position in wireframe-schema.json's oneOf array, and embedded-content-wireframe-update-schema.json to the first position in wireframe-update-mutation-schema.json's oneOf array, so they're tried before the generic shape/text/image/placeholder/webview cases that would otherwise shadow them. No schema field changes — ordering only.

### DIFF
--- a/lib/cjs/generated/mobileSessionReplay.d.ts
+++ b/lib/cjs/generated/mobileSessionReplay.d.ts
@@ -51,15 +51,23 @@ export type SlotSupportedCommonRecordSchema = CommonRecordSchema & {
 /**
  * Schema of a Wireframe type.
  */
-export type Wireframe = ShapeWireframe | TextWireframe | ImageWireframe | PlaceholderWireframe | WebviewWireframe | EmbeddedContentWireframe;
+export type Wireframe = EmbeddedContentWireframe | ShapeWireframe | TextWireframe | ImageWireframe | PlaceholderWireframe | WebviewWireframe;
 /**
- * Schema of all properties of a ShapeWireframe.
+ * Schema of all properties of an EmbeddedContentWireframe.
  */
-export type ShapeWireframe = CommonShapeWireframe & {
+export type EmbeddedContentWireframe = CommonShapeWireframe & {
     /**
      * The type of the wireframe.
      */
-    readonly type: 'shape';
+    readonly type: 'embedded_content';
+    /**
+     * Unique Id of the slot containing this embedded content.
+     */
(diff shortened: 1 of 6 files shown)


sethfowler-datadog:
> Generated Swift/Kotlin decoders try each oneOf case in declaration order and keep the first one that successfully decodes. SRShapeWireframe's required fields are a strict subset of SREmbeddedContentWireframe's, so with embedded-content listed last, an embedded-content payload was being matched and decoded as a generic shape wireframe first — silently dropping its required slotId and corrupting the type field. This broke any round-trip decode of an embedded-content wireframe (e.g. Flutter add-to-app placeholders), even though encoding was always correct.

@JuanNaranjoDD This seems like a bug in the decoders. This is a discriminated union; the `type` property tells you which variant you are dealing with. An `SRShapeWireframe` object has `type: "shape"`, and an `SREmbeddedContentWireframe` object has `type: "embedded_content"`, so the decoders should not confuse them. We should address the issue in the decoder implementations, or the problem will appear again in other situations. (Indeed, this may already be causing other bugs!)

sethfowler-datadog (changes requested):
See above; let's investigate the decoders before making this change.

gonzalezreal:
Hey @JuanNaranjoDD,

Thanks for looking into this. I’m going to close this PR and fix the issue in https://github.com/DataDog/dd-sdk-ios instead.

The underlying problem is in the generated decoders, not in the schema order. Wireframes are a discriminated union, so the `type` field should decide which concrete wireframe type gets decoded. Reordering the `oneOf` entries only changes which model wins under the current “try each type and keep the first one that decodes” strategy, which remains fragile and can break again whenever one wireframe schema is accepted as another.

PR #418, opened 2026-07-23
Merge Timeseries schema

barboraplasovska:
### What and why ?
Adds a new timeseries RUM event type to the schema, with two concrete event schemas: `timeseries-cpu-schema.json` and `timeseries-memory-schema.json`, registered in `rum-events-mobile-schema.json`.

This backs the new experimental mobile RUM Timeseries feature: periodic CPU and memory samples collected during a session, batched and flushed as their own event type.

### DIFF
--- a/schemas/rum-events-mobile-schema.json
+++ b/schemas/rum-events-mobile-schema.json
@@ -39,6 +39,17 @@
     },
     {
       "$ref": "telemetry-events-schema.json"
+    },
+    {
+      "title": "RumTimeseriesEvent",
+      "oneOf": [
+        {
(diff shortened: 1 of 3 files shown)


satween (approved):

Manu-Tran (approved):

bdibon (approved):
LGTM, I have a few suggestions.

**💬 suggestion:** ‏For reference it would be nice to add a sample JSON payload in the `samples/` directory.

**❓ question:** ‏There seem to be a common set of fields shared by all timeseries (id, name, schema, start, end). Would it be useful to factorize this into a schema reference `$ref`?

PR #427, opened 2026-08-12
[RUM-16985] Add browser WASM error metadata

ImaneLargou:
Adds the `browser+wasm` error source type and the `error.wasm_modules` array used to identify WebAssembly modules for stack trace symbolication. Includes generated definitions and a validating RUM event sample with a module URL and build ID.

Required by DataDog/browser-sdk#4920.


mormubis (approved):

jhssilva (approved):
👍

PR #438, opened 2026-08-26
[PANA-8797] Add string table clearing op to browser replay schema

sethfowler-datadog:
# Motivation

The browser session replay data format extracts strings from the individual replay operations and transmits them in a separate string table; this provides a number of benefits including cross-segment deduplication and improved compression. Today, we have no way to clear the string table; if it gets too big, we simply stop being able to add strings, and have to fall back to transmitting string literals embedded in the replay operations. It'd be better to be able to clear the string table; among other benefits, this would allow us to continue to benefit from deduplication in very large sessions.

# Changes

This PR adds a new browser session replay operation that clears the string table. It's a simple operation; it takes no arguments. Pretty straightforward!


gonzalezreal (approved):

JuanNaranjoDD (approved):
LGTM!

PR #440, opened 2026-09-03
Add image content changes to browser Session Replay

BeltranBulbarellaDD:
## What does this PR do?

Adds the browser Session Replay image-content change variant for visual resources such as captured canvas snapshots.

The compact replay encoding is:

```json
[15, [node_id, resource_id]]
```

The `resource_id` references an image uploaded separately as a hashed resource. This follows the existing mobile resource flow: the image bytes are sent through `POST /api/v2/replay` as a multipart resource part, while the normal replay stream carries only the lightweight node-to-resource reference.

## Scope

- Adds `image-content-change-schema.json`.
- Registers change type `15` in the browser replay change union.
- Exposes `ChangeType.ImageContent`.
- Updates browser replay samples and generated TypeScript outputs.

Canvas capture, resource deduplication/upload orchestration, and replay record generation in the Browser SDK are follow-up integration work.

## Consumer compatibility

The Browser SDK compatibility workflow regenerates its replay types from this PR. Its exhaustive replay decoder must add a `case 15` / `ChangeType.ImageContent` branch before that downstream check can pass; that consumer implementation belongs in the Browser SDK integration work.
Example https://github.com/DataDog/rum-events-format/pull/434

## Validation

- `yarn generate`
- `yarn build:cjs`
- `yarn build:esm`
- `yarn validate`
- `yarn format`


BeltranBulbarellaDD:
@codex review

gonzalezreal (approved):

schemas/session-replay/browser/changes/image-content-change-schema.json
```diff
@@ -0,0 +1,16 @@
...
+  "title": "ImageContentChange",
+  "description": "Browser-specific. Schema representing a change in an element's image content.",
+  "type": "array",
+  "minItems": 2,
+  "items": [
+    { "$ref": "node-references/node-id-schema.json" },
+    {
+      "type": "string",
+      "description": "The resource ID of the element's new image content."
+    }
```
sethfowler-datadog: Please use a `StringOrStringReference` here:

```suggestion
    { "$ref": "string-references/string-or-string-reference-schema.json" }
```

You should still be able to add a `description` property, I think.

You should also add a new string role to `string-role-schema.json`. Please use the name `StringResourceId`.

With these changes, the resource ids will be placed in the string table. That provides efficiency benefits, like cross-segment deduplication, and also adds support for the resource ids to our backend privacy tooling. (For example, if we want to mask all canvas content in a session when non-admin users view it, we can easily do that if the resource ids are in the string table and have a specific role we can look for.)

schemas/session-replay/browser/changes/image-content-change-schema.json
```diff
@@ -0,0 +1,16 @@
...
+  "$id": "session-replay/browser/changes/image-content-change-schema.json",
+  "title": "ImageContentChange",
+  "description": "Browser-specific. Schema representing a change in an element's image content.",
+  "type": "array",
+  "minItems": 2,
+  "items": [
+    { "$ref": "node-references/node-id-schema.json" },
+    {
+      "$ref": "string-references/string-or-string-reference-schema.json",
+      "description": "The resource ID of the element's new image content."
```
sethfowler-datadog: Doh! I know I suggested that you could add a "description", but if you look at the generated TypeScript code, it looks like doing that creates a duplicated version of `StringOrStringReference`. (And even though it duplicates it, it doesn't use the description you added here in the comment!) It'd probably be better to avoid that; maybe we can move this information into the "description" field of `ImageContentChange` itself?

sethfowler-datadog (approved):
Nice!

PR #445, opened 2026-09-10
RUM-18377 Remove local_cache_hit field from resource schema

barboraplasovska:
### What and why?
Removes the `local_cache_hit` boolean field from `schemas/rum/resource-schema.json`. It was a mobile-only signal for local-cache-hit detection, but the schema already has `delivery_type`/`transfer_size` fields and the backend's cache-status heuristic already derives local-cache and server-validated-cache hits from byte sizes for other platforms. Mobile is moving to populate those existing fields instead of maintaining a separate boolean.

### DIFF
--- a/lib/cjs/generated/rum.d.ts
+++ b/lib/cjs/generated/rum.d.ts
@@ -802,10 +802,6 @@ export type RumResourceEvent = CommonProperties & ActionChildProperties & ViewCo
          */
         readonly delivery_type?: 'cache' | 'navigational-prefetch' | 'other';
         /**
-         * Whether the resource was served from the device's local cache
-         */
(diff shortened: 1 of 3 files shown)


gonzalezreal (changes requested):
Moving SDKs to `delivery_type` and `transfer_size` makes sense, but we should not remove `local_cache_hit` while `_dd.format_version` is still 2.

Older SDK versions already send this field, so I suggest keeping it in the schema, marking it as deprecated, and removing it only in a future major format version.

gonzalezreal (approved):
🚀

bdibon (approved):

guillaumerob-dd (approved):

PR #446, opened 2026-09-14
Add boot_ntp field to mobile profile event schema

ambushwork:
Adds the `_dd.boot_ntp` field to the mobile profile event schema.

It carries the NTP-corrected device boot timestamp, in nanoseconds. Profiling samples are timestamped with boot time, so this lets the backend align them with absolute timestamps — the mobile equivalent of the browser's `_dd.clock_drift`.

Includes:
- Schema definition in `schemas/profiling/mobile/profile-event-schema.json`
- Validation allowlist entry in `scripts/validate.mjs`
- Regenerated TypeScript definitions


ambushwork:
@codex review

